PR #34, opened 2018-08-07
Https support for Keycloak installation

cborisa:
Enables HTTPS support for Keycloak installations when HTTPS_Proxy is not used. 

Lacks Keycloak_conn_validator due to issues with SSL truststore in puppet_http_connector (see Puppet ticket SERVER-1543)

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -37,6 +37,17 @@ class keycloak::config {
     mode   => '0750',
   }
 
+  if $keycloak::https {
+    #Copy keycloak keystore
+    file {'Keycloak keystore':
+      path   => "${keycloak::install_base}/standalone/configuration/${keycloak::ssl_keystore_file}",
+      owner  => $keycloak::user,
+      group  => $keycloak::group,
+      mode   => '0600',
+      source => "${keycloak::ssl_keystore_source}/${keycloak::ssl_keystore_file}"
+    }
+  }
+
   file { "${keycloak::install_base}/config.cli":
     ensure    => 'file',
     owner     => $keycloak::user,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -104,6 +104,21 @@
 # @param client_templates
 #   Hash that is used to define keycloak::client_template resources.
 #   Default is `{}`.
+# @param https 
+#   Defines if https should be configured.
+#   Default is false.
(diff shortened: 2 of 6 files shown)


manifests/init.pp
```diff
@@ -171,16 +192,18 @@
...
 
-  keycloak_conn_validator { 'keycloak':
-    keycloak_server => 'localhost',
-    keycloak_port   => $http_port,
-    use_ssl         => false,
-    timeout         => 60,
-    test_url        => '/auth/realms/master/.well-known/openid-configuration',
-    require         => Class['keycloak::service'],
+# Keycloak conn validator is active only for HTTP connection, as HTTPS connection works only with Puppet certs and not custom keystore (Ticket SERVER-1543)
+  if ! $keycloak::https {
```
treydock: Why not set `use_ssl => $keycloak::https` ?   Could also update port to point at either `$http_port` or `$https_port` , assuming SSL enabled support works.  Having `keycloak_conn_validator` is important for the custom types/providers.

manifests/config.pp
```diff
@@ -37,6 +37,17 @@
     mode   => '0750',
   }
 
+  if $keycloak::https {
```
treydock: Could the `java_ks` type be used here to build the keystore?  Similar to here: https://github.com/treydock/puppet-module-keycloak/blob/master/manifests/truststore/host.pp#L20.

manifests/init.pp
cborisa: I tried it, but as I put in description, Keycloak_conn_validator will always fails, as it tries to connect using Puppet node SSL certificate, and not custom SSL truststore used in Keycloak. 
At the moment according to the ticket mentioned this is not supported (it's an issue of Puppet::Network::HttpPool.http_instance )

treydock: Ah sorry missed that in description.  Using Puppet's HTTP stack was done because something copied from PuppetDB module, maybe at some point need to revisit and use standard `net/http` or some other HTTP communication stack.

treydock:
Besides comment on potentially using `java_ks`, things look good.  Will need to rebase against master and squash commits.  I updated master to fix acceptance tests in travis-ci.

treydock:
This is no longer relevant with Keycloak 17+ using Quarkus

PR #50, opened 2019-02-11
Add option for service environment file

asieraguado:
First, thanks for this great module! We're using the it for setting up a Keycloak instance that uses a custom extension. This extension needs configuring some environment variables, and the best seems to manage them in the `keycloak.service` file.

The pull request adds a parameter for including the `EnvironmentFile` option into the systemd service file, so it makes possible to configure environment variables for the Keycloak service.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -182,7 +182,7 @@ class keycloak (
   Boolean $manage_sssd_config = true,
   Array $sssd_ifp_user_attributes = [],
   Boolean $restart_sssd = true,
-  Optional[String] $service_environment_file = undef
+  Optional[Stdlib::Absolutepath] $service_environment_file = undef
 ) inherits keycloak::params {
 
   $download_url = pick($package_url, "https://downloads.jboss.org/keycloak/${version}/keycloak-${version}.tar.gz")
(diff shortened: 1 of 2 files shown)


templates/keycloak.service.erb
```diff
@@ -5,6 +5,9 @@ After=network.target
 [Service]
 Type=idle
 #Environment="JAVA_OPTS=-Xms1024m -Xmx20480m -XX:MaxPermSize=768m"
+<% if !scope['keycloak::service_environment_file'].nil? -%>
```
treydock: This could be simplified with just `<% if scope['keycloak::service_environment_file'] -%>`.

manifests/init.pp
```diff
@@ -180,6 +182,7 @@
   Boolean $manage_sssd_config = true,
   Array $sssd_ifp_user_attributes = [],
   Boolean $restart_sssd = true,
+  Optional[String] $service_environment_file = undef
```
treydock: Should probably be `Optional[Stdlib::Absolutepath]`.

asieraguado:
Updated the PR

treydock:
@asieraguado Thanks for the contribution!

treydock:
Released in 3.4.0

Issue #49, opened 2019-02-11
`service_java_opts` parameter not actually being used


danifr:
I'm trying to extend `$JAVA_OPTS` via the class param `service_java_opts` but it does not seem to work. 

I took a look at the code and it seems is defined but never used.
https://github.com/treydock/puppet-module-keycloak/search?q=service_java_opts&unscoped_q=service_java_opts

Thanks a lot.
Daniel.

treydock:
Released in 3.4.0

PR #51, opened 2019-02-13
JAVA_OPTS via systemd unit Environment variable

danifr:
Fixes https://github.com/treydock/puppet-module-keycloak/issues/49

### DIFF
--- a/templates/keycloak.service.erb
+++ b/templates/keycloak.service.erb
@@ -8,7 +8,11 @@ Type=idle
 EnvironmentFile=<%= scope['keycloak::service_environment_file'] %>
 <% end -%>
 <% if scope['keycloak::service_java_opts'] -%>
-Environment="JAVA_OPTS=<%= scope['keycloak::service_java_opts'].join(' ') %>"
+<% if scope['keycloak::server_java_opts'].is_a?(Array) -%>
+Environment="JAVA_OPTS=$JAVA_OPTS <%= scope['keycloak::service_java_opts'].join(' ') %>"
+<% else %>
+Environment="JAVA_OPTS=$JAVA_OPTS <%= scope['keycloak::service_java_opts'] %>"
+<% end -%>
 <% end -%>
 User=<%= scope['keycloak::user'] %>
 Group=<%= scope['keycloak::group'] %>
--- a/templates/keycloak.service.erb
+++ b/templates/keycloak.service.erb
@@ -4,9 +4,11 @@ After=network.target
 
 [Service]
 Type=idle
-#Environment="JAVA_OPTS=-Xms1024m -Xmx20480m -XX:MaxPermSize=768m"
 <% if scope['keycloak::service_environment_file'] -%>
 EnvironmentFile=<%= scope['keycloak::service_environment_file'] %>
+<% end -%>
+<% if scope['keycloak::service_java_opts'] -%>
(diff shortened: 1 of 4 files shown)


treydock:
@danifr Would need to update `$service_java_opts` in `keycloak::params` to match old hardcoded values.  The values in `keycloak::params` are different than previous hardcoded values.

danifr:
Are you refereeing to these values? https://github.com/treydock/puppet-module-keycloak/blob/master/templates/keycloak.service.erb#L7

I was thinking that I can do something similar to what Asier is doing in the other PR. Leave `$service_java_opts` undef (so JAVA_OPTS will take keycloak's installation default values) and only add it to the systemd unit Environment if the variable is redefined by the user.

What do you think?

treydock:
Ya, having default be `undef` (remove current values from params.pp) and only setting if defined is fine.  I forgot the current file has the values commented out so default being `undef` maintains backwards compatibility.

danifr:
My only doubt now is if:
```
<% if scope['keycloak::service_java_opts'] -%>
Environment="JAVA_OPTS=<%= scope['keycloak::service_java_opts'].join(' ') %>"
<% end -%>
```
should become (concat instead of reassigning):
```
<% if scope['keycloak::service_java_opts'] -%>
Environment="JAVA_OPTS=$JAVA_OPTS <%= scope['keycloak::service_java_opts'].join(' ') %>"
<% end -%>
```

templates/keycloak.service.erb
```diff
@@ -4,10 +4,12 @@ After=network.target
 
 [Service]
 Type=idle
-#Environment="JAVA_OPTS=-Xms1024m -Xmx20480m -XX:MaxPermSize=768m"
 <% if scope['keycloak::service_environment_file'] -%>
 EnvironmentFile=<%= scope['keycloak::service_environment_file'] %>
 <% end -%>
+<% if scope['keycloak::service_java_opts'] -%>
+Environment="JAVA_OPTS=<%= scope['keycloak::service_java_opts'].join(' ') %>"
```
treydock: Need to check if `String` vs `Array` and only use `join` for `Array`:

```
2.4.0 :001 > '-Xmx=1024m'.join(' ')
NoMethodError: undefined method `join' for "-Xmx=1024m":String
```

Could probably do second condition inside current `if` that is `if scope['keycloak::server_java_opts'].is_a?(Array)` and the `else` condition would be for `String`.

treydock:
I think concat would be good as would allow `EnvironmentFile` from #50 to also set `JAVA_OPTS`.

templates/keycloak.service.erb
```diff
@@ -4,10 +4,16 @@ After=network.target
 
 [Service]
 Type=idle
-#Environment="JAVA_OPTS=-Xms1024m -Xmx20480m -XX:MaxPermSize=768m"
 <% if scope['keycloak::service_environment_file'] -%>
 EnvironmentFile=<%= scope['keycloak::service_environment_file'] %>
 <% end -%>
+<% if scope['keycloak::service_java_opts'] -%>
+  <% if scope['keycloak::server_java_opts'].is_a?(Array) -%>
```
treydock: While the indentation is more readable, it will add unnecessary white space to the unit file.  I'd recommend removing white space before the condition portions of inner `if` statement.

treydock:
Released in 3.4.0

PR #53, opened 2019-04-05
manage user support

cborisa:
Ability to control if the module should be control the Linux user for keycloak.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -36,6 +36,6 @@ class keycloak::install {
     creates         => "${keycloak::install_dir}/keycloak-${keycloak::version}/bin",
     cleanup         => true,
     user            => $keycloak::user,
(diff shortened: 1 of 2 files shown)


manifests/install.pp
```diff
@@ -35,8 +35,7 @@
     source          => $keycloak::download_url,
     creates         => "${keycloak::install_dir}/keycloak-${keycloak::version}/bin",
     cleanup         => true,
-    user            => 'keycloak',
-    group           => 'keycloak',
+    user            => $keycloak::user,
+    group           => $keycloak::user,
```
treydock: Should probably be `$keycloak::group` for `group`

cborisa: You are right. FIxed it already

treydock:
Released to forge as 3.5.0

PR #59, opened 2019-05-17
Postgresql support

verrydtj:


files/database/postgresql/module.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" ?>
```
treydock: I don't think this file is necessary, looks like your using the template and not this file.

treydock:
Would be good to add example of how you call `keycloak` class to add postgresql support, like example of postgresql specific parameters necessary to add postgresql support.  Just add it under existing mysql example.

verrydtj:
Indeed, I will add this next week and also remove that old module.xml file.?

________________________________

treydock:
Released as 3.7.0

PR #79, opened 2019-08-29
Fix #78. Add clustered mode support

### LINKED ISSUE #78: Add support for standalone clustered mode
Info about this mode: https://www.keycloak.org/docs/latest/server_installation/index.html#_standalone-ha-mode

I can prepare a PR for this. If I'm not mistaken, I think it is just a matter of changing the operating mode (standalone.xml or standalone-ha.xml) for the ExecStart in  `keycloak.service.erb`.

Am I missing something? Not sure if this
https://github.com/treydock/puppet-module-keycloak/blob/master/templates/config.cli.erb needs to be updated too.

danifr:
According to: https://www.keycloak.org/docs/latest/server_installation/index.html#standalone-clustered-boot-script

> You use the same boot scripts to start Keycloak as you do in standalone mode. The difference is that you pass in an additional flag to point to the HA config file
> 
> To boot the server:
> $ .../bin/standalone.sh --server-config=standalone-ha.xml

Those were the only changes needed. About the tests, I would need your help for a more meaningful ones.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -163,6 +163,8 @@
 #   Boolean that determines if SSSD should be restarted
 # @param service_environment_file
 #   Path to the file with environment variables for the systemd service
+# @param operating_mode
+#   Keycloak operating mode deployment
 #
 class keycloak (
   String $version               = '6.0.1',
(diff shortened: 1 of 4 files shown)


treydock:
Also need to update `config.cli.erb` to use `embed-server --server-config=standalone-ha.xml` otherwise that script will end up modifying `standalone.xml` and not `standalone-ha.xml`. The use of a file like config.cli isn't well documented by Keycloak so not surprising it's not mentioned for clustered mode.

treydock:
Released in v5.3.0

PR #83, opened 2019-09-04
Vagrant: add Ubuntu 1804 box

mattock:
Provisioning was split into OS-specific and OS-agnostic parts to reduce code
duplication.


mattock:
I tested both CentOS 7 and Ubuntu 18.04 and in both cases bringing up the VM and provisioning it completed correctly. That is, Keycloak UI loaded just fine.

treydock:
@mattock Thanks for the contribution.

PR #85, opened 2019-09-25
Enable defining bind address for the Keycloak systemd service

mattock:
This is useful for example when there is a reverse proxy is in front of Keycloak


treydock:
Need to update `.fixutres.yml` to use stdlib `4.25.0` and update `metadata.json` minimum version of stdlib to match.

treydock:
Will need to rebase against master branch since #86 was merged.

mattock:
Rebased and Travis seems happy now as well.

treydock:
Released in v5.5.0

PR #86, opened 2019-09-25
Enable passing extra options to Keycloak in the systemd unit file

mattock:
This allows passing arbitrary command-line options as a string to standalone.sh in the systemd unit file.


treydock:
Released in v5.5.0

Issue #92, opened 2019-10-28
Question about support for MySQL8


danifr:
Hi,

I'm playing around with Keycloak on MySQL8. I followed this blogpost https://pedrohosilva.wordpress.com/2019/08/11/04-configurando-banco-de-dados-mysql/ and I can confirm that it works.

Now, I'm trying to puppetize it but it did not work straight away. 

I provided the module with a custom `datasource_jar_source` pointing it to the new mysql-connector-java-8.0.XX.jar, but the service won't start.

After some investigation, I realized that this is the line that breaks it https://github.com/treydock/puppet-module-keycloak/blob/master/templates/config.cli.erb#L28

If I manually delete it from the standalone file, and restart the service, keycloak service starts OK

I don't understand very well the logic behind it.

`if (outcome != success)  .... `

Does the outcome always != success in my case?
Is there a simple way I can avoid that line being added to the standalone file?

Thanks a lot,
Daniel.

treydock:
What does your Puppet code or manifest look like that is using this module?  The string that contains `(outcome != success)` is part of the config.cli that uses JBoss CLI to write out standalone.xml. The part you're referring to writes out part of the MySQL configs in standalone.xml if a certain XML element does not exist.

danifr:
It looks like:
```
  class { 'keycloak':
    version                  => $version,
    proxy_https              => true,
    manage_datasource        => false,
    operating_mode           => $keycloak_operating_mode,
    datasource_driver        => $db_driver,
    datasource_host          => $db_host,
    datasource_port          => $db_port,
    datasource_dbname        => $db,
    datasource_username      => $db_username,
    datasource_password      => "$password,
    datasource_jar_source    => "${keycloak_config_dir}/mysql-connector-java-8.0.18.jar",
    service_environment_file => "${keycloak_base_dir}/keycloak.env",
  }
```

treydock:
What does `/opt/keycloak/config.cli` look like for you?

The part that appears to be causing problems for you is setting the MySQL driver class name to `com.mysql.jdbc.jdbc2.optional.MysqlXADataSource`. Maybe this isn't valid for MySQL 8? The version of MySQL I've tested this with is 5.6 that ships with RHEL7 and MariaDB that ships with RHEL 8, and whatever version ships with Debian 9 and Ubuntu 18.04. 

This is what it looks like in standalone.xml or in your case would be standalone-ha.xml:

```
[...]

PR #93, opened 2019-10-28
Make JDBC xa-datasource-class name configurable

danifr:
As discussed here: https://github.com/treydock/puppet-module-keycloak/issues/92

This is the first approach I came up with maybe we can discuss a bit further.

When this gets accepted I might opened a new PR to sort the variables alphabetically :)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -96,6 +96,8 @@
 #   This parameter is required if `datasource_driver` is `oracle`.
 # @param datasource_module_source
 #   Source for datasource module.xml. Default depends on `datasource_driver`.
+# @param datasource_xa_class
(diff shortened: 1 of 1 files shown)


treydock:
I'd prefer to keep the parameters to just be `datasource_` and thus not mysql or postgresql specific. So I think the one that is missing is `datasource_class` which maybe confusing given Puppet has classes so calling it `datasource_java_class` would likely be better.

The `*_jar_source` parameters already exist as `datasource_jar_source` and the `pick()` is performed in the classes like `keycloak::datasource::mysql`.

danifr:
I think I addressed all your comments. Thanks.

treydock:
@danifr Thanks for the contribution, this is released in v5.7.0

Issue #94, opened 2019-11-19
Support for specific technology preview features


danifr:
My idea is to create `profile.properties` file and have there the list of enabled features.

https://www.keycloak.org/docs/latest/server_installation/#profiles

I'll add a PR later today. We can discuss the implementation details

treydock:
Closing via #95

PR #95, opened 2019-11-19
Add option to enable tech preview features

danifr:
https://github.com/treydock/puppet-module-keycloak/issues/94

Probably worth to add some tests for this too.
Not sure if I should also support a way to disable/enable per feature.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -37,6 +37,17 @@ class keycloak::config {
     mode   => '0750',
   }
 
+  if $keycloak::tech_preview_features {
+    file { "${keycloak::install_base}/standalone/configuration/profile.properties":
+      ensure  => 'file',
+      owner   => $keycloak::user,
+      group   => $keycloak::group,
+      content => template('keycloak/profile.properties.erb'),
+      mode    => '0750',
+      require => File["${keycloak::install_base}/standalone/configuration"],
+    }
+  }
+
   file { "${keycloak::install_base}/config.cli":
     ensure    => 'file',
     owner     => $keycloak::user,
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -37,15 +37,14 @@ class keycloak::config {
     mode   => '0750',
   }
 
-  if $keycloak::tech_preview_features {
-    file { "${keycloak::install_base}/standalone/configuration/profile.properties":
-      ensure  => 'file',
-      owner   => $keycloak::user,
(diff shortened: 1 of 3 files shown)


manifests/config.pp
```diff
@@ -37,6 +37,17 @@
...
 
+  if $keycloak::tech_preview_features {
+    file { "${keycloak::install_base}/standalone/configuration/profile.properties":
+      ensure  => 'file',
+      owner   => $keycloak::user,
+      group   => $keycloak::group,
+      content => template('keycloak/profile.properties.erb'),
+      mode    => '0750',
+      require => File["${keycloak::install_base}/standalone/configuration"],
+    }
```
treydock: Would be good to `notify` the Keycloak service class, `Class['keycloak::service']`. Also to allow undoing this change maybe set `ensure` to `$profile_properties_ensure` and in `init.pp` set the value based on presence of `tech_preview_features`.  This way if someone no longer set `tech_preview_features` the file will get deleted and Keycloak restarted.

manifests/init.pp
```diff
@@ -250,12 +252,19 @@
   Boolean $restart_sssd = true,
   Optional[Stdlib::Absolutepath] $service_environment_file = undef,
   Enum['standalone', 'clustered'] $operating_mode = 'standalone',
+  Optional[Array] $tech_preview_features = undef,
```
treydock: Make this just `Array` with empty array as default.

manifests/init.pp
```diff
@@ -250,12 +252,19 @@
...
 ) {
 
   if ! $facts['os']['family'] in ['RedHat','Debian'] {
     fail("Unsupported osfamily: ${facts['os']['family']}, module ${module_name} only support osfamilies Debian and Redhat")
   }
 
+  $profile_properties_ensure = $tech_preview_features ? {
```
treydock: If this becomes just empty array can do something like `if empty($tech_preview_features) { `

templates/profile.properties.erb
```diff
@@ -0,0 +1,5 @@
+<% if scope['keycloak::tech_preview_features'] -%>
```
treydock: Once `tech_preview_features` is only an array, can remove this if statement.

danifr:
@treydock sorry I've been busy and I couldn't address your comments...
Thanks for taking the over!

treydock:
This was released in v5.8.0

PR #97, opened 2019-12-05
Add support for managing client attributes

mattock:
Due to the way provider is implemented it is only possible to add attributes and
modify their values, not to get rid of them once they're set (manually or with Puppet).

### DIFF
--- a/lib/puppet/type/keycloak_client.rb
+++ b/lib/puppet/type/keycloak_client.rb
@@ -120,6 +120,13 @@ Manage Keycloak clients
     defaultto []
   end
 
+  newproperty(:attributes) do
+    desc 'attributes'
+    validate do |attributes|
+      raise('Property attributes must be a hash') unless attributes.is_a?(Hash)
+    end
+  end
+
   autorequire(:keycloak_client_scope) do
     requires = []
     catalog.resources.each do |resource|
--- a/spec/unit/puppet/type/keycloak_client_spec.rb
+++ b/spec/unit/puppet/type/keycloak_client_spec.rb
@@ -142,6 +142,22 @@ describe Puppet::Type.type(:keycloak_client) do
     end
   end
 
+  describe 'hash properties' do
+    # Hash properties
+    [
+      :attributes,
+    ].each do |p|
+      it "should accept hash for #{p}" do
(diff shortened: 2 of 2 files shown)


mattock:
@treydock with the current provider implementation it seem not to be possible to remove keys from the attributes hash. Even when Puppet puts _"attributes" : {}_ to the JSON file, kcadm will not remove any pre-existing keys from the client's attributes. It seems that using a different kcadm incantation for managing attributes would be necessary to manage the attributes hash as a whole. My question is: is that desirable and worth the effort, in your opinion?

Our use-case for this is - for now - to be able to set the login theme on for our clients.

treydock:
Look at how a provider like the one for `keycloak_client_scope`. Rather than having a Hash for attributes each value in `attributes` is added as a property to the type. Are `attributes` for a client setting specific keys that Keycloak understands or is it a freeform Hash that can have any key?  If your setting specific keys that keycloak knows how to process then those can each be new properties.

https://github.com/treydock/puppet-module-keycloak/blob/master/lib/puppet/type/keycloak_client_scope.rb#L47

https://github.com/treydock/puppet-module-keycloak/blob/master/lib/puppet/provider/keycloak_client_scope/kcadm.rb#L28-L29

https://github.com/treydock/puppet-module-keycloak/blob/master/lib/puppet/provider/keycloak_client_scope/kcadm.rb#L54-L55

I don't know if setting the value to `nil` or `null` if not defined will remove from Keycloak.

treydock:
I found that if you define the login theme in web UI it sets attributes to have `login_theme` but then you can keep all the other default attributes and remove `login_theme` and it goes away. There are just some attributes that seem to not be something that can be deleted even when attributes is empty.  I verified setting `login_theme` to `null` in JSON removes it from attributes.

treydock:
@mattock Check out #98 where I added the `login_theme` property. Are there any other things under `attributes` you would need?

treydock:
Closing in favor of #98. If more client attributes are needed, open an issue with the attributes needed and their default values and I can make the necessary changes.

mattock:
At the moment we're only interested in the login_theme attribute. I'll test your implementation. Thanks!

treydock:
This was released in v5.8.0

PR #107, opened 2019-12-19
Fix problem with keycloak admin user creation

mattock:
We've encountered this problem on several Keycloak versions (4.8.3 and up). It seems that this is a [known bug](https://issues.jboss.org/browse/KEYCLOAK-9923). [A fix](https://github.com/keycloak/keycloak/pull/5978) has been merged but it won't help with older Keycloak versions.

This PR adds the documented workaround to the Keycloak admin user creation Exec.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -25,9 +25,10 @@ class keycloak::config {
   $_add_user_keycloak_args = "--user ${keycloak::admin_user} --password ${keycloak::admin_user_password} --realm master"
   $_add_user_keycloak_state = "${keycloak::install_base}/.create-keycloak-admin-${keycloak::datasource_driver}"
   exec { 'create-keycloak-admin':
-    command => "${_add_user_keycloak_cmd} ${_add_user_keycloak_args} && touch ${_add_user_keycloak_state}",
-    creates => $_add_user_keycloak_state,
-    notify  => Class['keycloak::service'],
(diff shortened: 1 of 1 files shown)


treydock:
Looks like some operating systems don't recognize the flag you added:

```
Notice: /Stage[main]/Keycloak::Config/Exec[create-keycloak-admin]/returns: Unrecognized option: --add-modules
  Notice: /Stage[main]/Keycloak::Config/Exec[create-keycloak-admin]/returns: Error: Could not create the Java Virtual Machine.
  Notice: /Stage[main]/Keycloak::Config/Exec[create-keycloak-admin]/returns: Error: A fatal exception has occurred. Program will exit.
```

treydock:
Also this module hasn't been tested with anything before 6.0.1 for a while and this issue never cropped up in the acceptance tests that runs the Exec. Are you still on a version of Keycloak before 6.0.1?

mattock:
The installation with the problem is Keycloak 4.8.3.Final. I can no longer remember if the Keycloak 8.x instance I set up a couple of days ago had this problem or not.

I'm perfectly ok with this PR getting closed: I can working around this problem in a profile or patch Keycloak to fix the problem until we are able to upgrade.

treydock:
If you can figure out a way to make this change compatible with all operating systems this module supports, that's fine. Seems like might not be worth the effort so I'll close this out but can re-open if you need this change and can make it work for all supported operating systems.

PR #108, opened 2019-12-24
Add two new realm properties

mattock:
This brings with it unit testing of integer-valued properties


treydock:
Released in v6.1.0

PR #109, opened 2019-12-25
Add support for access.token.lifespan client attribute

mattock:
This also fixes handling of attributes that have dots in their names, which
means most of them, excluding "login_theme". Here is a truncated example:

```
  "attributes" : {
    "saml.assertion.signature" : "false",
    "access.token.lifespan" : "600",
    "saml.multivalued.roles" : "false",
    "saml.force.post.binding" : "false",
    "saml.encrypt" : "false",
    "login_theme" : "keycloak",
    "saml.server.signature" : "false",
    "saml.server.signature.keyinfo.ext" : "false",
     --- snip ---
```
Without the dot_key fix in the "flush" method Puppet added and tried to manage _access_token_lifespan_ instead of _access.token.lifespan_.

### DIFF
--- a/lib/puppet/provider/keycloak_client/kcadm.rb
+++ b/lib/puppet/provider/keycloak_client/kcadm.rb
@@ -8,7 +8,22 @@ Puppet::Type.type(:keycloak_client).provide(:kcadm, parent: Puppet::Provider::Ke
   def attributes_properties
     [
       :login_theme,
+      :access_token_lifespan,
     ]
+  end
+
+  def dot_attributes_properties
+    [
+      :access_token_lifespan,
+    ]
+  end
+
+  def attribute_key(property)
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/keycloak_client/kcadm.rb
```diff
@@ -222,7 +229,12 @@ def flush
           unless data.key?(:attributes)
             data[:attributes] = {}
           end
-          data[:attributes][property] = value
+          if dot_attributes_properties.include?(property)
```
treydock: Something like this will also be needed in the `create` method.

mattock: Of course. I'm on it.

treydock:
Released in v6.1.0

PR #127, opened 2020-03-11
Allow enabling/disabling client authorization services

mattock:
While this is a seemingly trivial change, there are issues in the Keycloak API (at least in 6.x) that makes the user-experience less than optimal and require discussion.

First of all _"authorizationServicesEnabled": true_ depends on _"serviceAccountsEnabled": true_. Trying to set the latter to true while former is false will give an 500 Internal Server Error. We may be able to prevent such transitions with proper validation.

Secondly  _"authorizationServicesEnabled"_ from _true_ to _false_ works. But changing it back from _false_ to _true_ does not - you get 500 Internal server error and this in the logs:

```
2020-03-11 09:44:30,247 ERROR [org.keycloak.services.error.KeycloakErrorHandler] (default task-26) Uncaught server error: java.lang.RuntimeException: Client does not have a service account.
```
This despite the fact that _"serviceAccountsEnabled": true_. This is how the keycloak_client provider currently tries to make this change:
```
{
     "clientId": "www.foobar.net",
     "authorizationServicesEnabled": true
 }
```

That does not work, probably due to reasons internal to Keycloak. What would work is this:

```
{
     "clientId": "www.foobar.net",
     "serviceAccountsEnabled": true,
     "authorizationServicesEnabled": true
 }
```
So, for this particular transition _serviceAccountsEnabled_ has to be explicitly defined, even if the value of serviceAccountsEnabled does not change.

There is also a minor issue. Because _"kcmadm-wrapper.sh get clients/\<client-id\>"_  does not display _authorizationServicesEnabled_ at all if it is not defined (i.e. is false), having _"authorization_services_enabled => false"_ makes Puppet run kcadm-wrapper.sh on every run. While it does not trigger failures or change anything it does pollute the Puppet reports. Setting the property to undefined is ok - no output is produced.

@treydock any preferences on how to handle this? Fix what can be fixed easily (e.g. add more validation) and document the rest? Or try to fix all of this?

### DIFF
--- a/lib/puppet/provider/keycloak_client/kcadm.rb
+++ b/lib/puppet/provider/keycloak_client/kcadm.rb
@@ -243,6 +243,12 @@ Puppet::Type.type(:keycloak_client).provide(:kcadm, parent: Puppet::Provider::Ke
         end
       end
 
+      # Keycload API requires "serviceAccountsEnabled": true to be present in
+      # the JSON when "authorizationServicesEnabled": true
+      if data['authorizationServicesEnabled'] && data['serviceAccountsEnabled'].nil?
+        data[:serviceAccountsEnabled] = true
+      end
+
       # Only update if more than clientId set
       if data.keys.size > 1
         t = Tempfile.new('keycloak_client')
--- a/lib/puppet/type/keycloak_client.rb
+++ b/lib/puppet/type/keycloak_client.rb
@@ -122,6 +122,26 @@ Manage Keycloak clients
     defaultto :false
   end
 
+  newproperty(:authorization_services_enabled, boolean: true) do
+    desc 'authorizationServicesEnabled'
+    newvalues(:true, :false)
+    defaultto :false
+
+    # If authorizationServicesEnabled is set to false it will not be present in
+    # "get client/<clientname>" output. Puppet will thus see it as "absent".
+    # This custom insync? implementation prevents Puppet from trying to set
+    # the property to false on every run.
+    def insync?(is)
+      if is == :true && resource[:authorization_services_enabled] == :true
+        true
+      elsif is == :absent && resource[:authorization_services_enabled] == :false
+        true
+      else
+        false
+      end
+    end
+  end
+
   newproperty(:public_client, boolean: true) do
     desc 'enabled'
     newvalues(:true, :false)
@@ -181,6 +201,12 @@ Manage Keycloak clients
       end
     end
     requires
+  end
+
+  validate do
+    if self[:authorization_services_enabled] == :true && self[:service_accounts_enabled] == :false
+      raise "Keycloak_client[#{self[:name]}] must have service_accounts_enabled => true if authorization_services_enabled => true"
+    end
   end
 
   def self.title_patterns
--- a/spec/acceptance/5_client_spec.rb
+++ b/spec/acceptance/5_client_spec.rb
@@ -10,12 +10,14 @@ describe 'keycloak_client define:', if: RSpec.configuration.keycloak_full do
       }
       keycloak_realm { 'test': ensure => 'present' }
       keycloak_client { 'test.foo.bar':
-        realm                 => 'test',
-        root_url              => 'https://test.foo.bar',
-        redirect_uris         => ['https://test.foo.bar/test1'],
-        default_client_scopes => ['address'],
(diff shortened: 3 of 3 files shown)


treydock:
If two properties depend on the value of another you'll need to update validation in type, example : https://github.com/treydock/puppet-module-keycloak/blob/master/lib/puppet/type/keycloak_flow.rb#L168

As long as the properties have a default value, they will always be in the API JSON even if not defined by puppet resource.

Will need some tests for adding value on create and modify here: https://github.com/treydock/puppet-module-keycloak/blob/master/spec/acceptance/5_client_spec.rb

Also unit tests for the new properties and any validations performed.

Does managing the service account make sense to avoid the error you posted? I don't know if that would be via client API or a different API call.

mattock:
Thanks for the pointers. I will look into fixing the issues.

Keycloak admin REST API seems rather devoid of service-account -related API calls, so I need to look into that separately.

mattock:
I fixed some of the issues and will continue later. I will squash the commits as the last step.

lib/puppet/type/keycloak_client.rb
```diff
@@ -183,6 +189,12 @@ def should_to_s(_newvalue)
     requires
   end
 
+  validate do
+    if self[:authorization_services_enabled] == :true && self[:service_accounts_enabled] == :false
+      raise "Keycloak_client[#{self[:name]}] must have service_accounts_enabled => true if authorization_services_enabled => true"
+     end
```
treydock, line 195: Currently rubocop is failing:

```
lib/puppet/type/keycloak_client.rb:195:6: W: Lint/EndAlignment: end at 195, 5 is not aligned with if at 193, 4.
     end
     ^^^
```

mattock:
I fixed the problem where Puppet tries to set authorization_services_enabled to false on every run. I also added acceptance tests for it (create + change). The acceptance tests now pass:

```
Finished in 5 minutes 27 seconds (files took 2 minutes 27.6 seconds to load)
6 examples, 0 failures
```

As far as I can see the only thing that requires resolving is the transitioning _authorization_services_enable_ from _false_ to _true_ when _service_accounts_enabled_ is already set to _true_. This is the issue where Keycloak spits out an internal server error.

I'll check the Travis errors to see if they're related to my work.

mattock:
I fixed one issue with Travis that was about incompatibility between how booleans are checked and the extra type-level validation I added for authorization_services_enabled and service_accounts_enabled. The former can't be checked standalone. That made two more tests pass.

The remaining two Travis failures don't look related to my work as they're all in spec/classes/init_spec.rb.

treydock:
The issues were because Facter 4 was getting installed which breaks a lot of things. I am fixing in #133

treydock:
Rebase against master branch and you should get a fix that will make unit tests no longer break

mattock:
The provider now ensures that when "authorizationServicesEnabled" is transitioned from false to true the JSON will get "serviceAccountsEnabled": true. As mentioned above, without it Keycloak will error out.

I squashed the commits as well.

mattock:
@treydock from my perspective this PR is ready. What do you think?

lib/puppet/type/keycloak_client.rb
```diff
@@ -183,6 +203,12 @@ def should_to_s(_newvalue)
     requires
   end
 
+  validate do
+    if self[:authorization_services_enabled] == :true && self[:service_accounts_enabled] == :false
```
treydock, line 207:
```suggestion
    if self[:authorization_services_enabled] == :true && self[:service_accounts_enabled] != :true
```

lib/puppet/provider/keycloak_client/kcadm.rb
```diff
@@ -243,6 +243,12 @@ def flush
         end
       end
 
+      # Keycload API requires "serviceAccountsEnabled": true to be present in
+      # the JSON when "authorizationServicesEnabled": true
+      if data['authorizationServicesEnabled'] && data['serviceAccountsEnabled'].nil?
+        data[:serviceAccountsEnabled] = true
+      end
+
```
treydock, line 251:
```suggestion
```

treydock:
I had some suggested changes but after closer inspection things look good.

mattock:
@treydock thanks!

PR #131, opened 2020-04-23
Add support for defining smtpServer from realms

mattock:


mattock:
I wanted to put this here for discussion before wrapping this up (spec, docs). So, the trivial change (commit b6150fb) does work correctly. Say, I have this Puppet code:

```
keycloak_realm { 'foobar':
  ensure => 'present',
  smtp_server => { 'password' => 'foobar',
                   'starttls' => 'true',
                   'auth'     => 'true',
                   'port'     => '587',
                   'host'     => 'localhost',
                   'replyTo'  => 'john@example.org',
                   'from'     => 'john@example.org',
                   'ssl'      => '',
                   'user'     => 'john', }                      
}
```
The result is the same as if I had configured the same things manually.

That said, based on practical tests the Keycloak API does zero validation on what gets passed to it as smtpServer parameters. You can pass it non-existing parameters ("broken" => "true") or leave out mandatory parameters (Host, From, Username, Password). This is probably because smtpServer is a generic Map datatype (see [RealmRepresentation docs](https://www.keycloak.org/docs-api/6.0/rest-api/index.html#_realmrepresentation)). Enforcement of parameters is only done at the GUI level afaics.

There's also another thing. The API returns SMTP password as '*******'. This will trigger a "change" on every Puppet run. I added a second commit (4683e7e) that makes Puppet ignore changes to the password, though I'm not sure which one is preferable:
* "Changes" on every run
* "Do not change at all after initially setting the password" (but notify the user about it)

Thoughts?

treydock:
For handling passwords, you have to do something like this: https://github.com/treydock/puppet-module-keycloak/blob/master/lib/puppet/type/keycloak_ldap_user_provider.rb#L115-L142

The issue with the approach taken from the link is that the parameter you've added is a Hash and so doing insync check is completely different. You will have to built new hashes that omit password and compare those two hashes. If password is defined be good to emit warning like is done for LDAP bind credentials.

Also for validation you can use `validate` block and here's an example: https://github.com/sensu/sensu-puppet/blob/master/lib/puppet/type/sensu_ldap_auth.rb#L52

You have to validate the keys and check that the keys provided are valid and also check for required keys.  That example also has example of filtering out password from logs when dealing with a Hash.

Lastly one other approach if it makes things easier is instead of having `smtp_server` property take each key and make it a property like `smtp_server_password` and `smtp_server_host` that would make validation much simpler and also make the `insync?` check for password easier to deal with.  Then in the provider you just build the hash like so:

```
smtp_server = {}
smtp_server['host'] = resource[:smtp_server_host] if resource[:smtp_server_host]
...
# add smtp_server hash to data passed to API
```

mattock:
I considered handling the SMTP settings individually but wanted your feedback first. In fact, I think that approach makes most sense given that there's no validation of the smtpServer map at the Keycloak end. Way fewer chances for the user of the module to shoot him/herself in the foot. I'll investigate that route next.

mattock:
Here the new implementation. I could not run "pdk test unit" so unit tests might fail - I need to fix that for the authorizationServicesEnabled PR anyways. "pdk validate" passes for my changes and I was unable to break this with my manual tests.

lib/puppet/provider/keycloak_realm/kcadm.rb
```diff
@@ -123,11 +143,15 @@ def create
       if self.class.browser_security_headers.include?(property) && !data.key?('browserSecurityHeaders')
         data['browserSecurityHeaders'] = {}
       end
+      if self.class.smtp_server_properties.include?(property)
```
treydock, line 146: This should be `self.class.smtp_server_properties.include?(property) && !data.key?('smtpServer')` and then if this is true it adds the hash `data['smtpServer'] = {}` so that you are only adding the key if SMTP server settings exist. Then this current assignemtn gets moved to line 155.

lib/puppet/provider/keycloak_realm/kcadm.rb
```diff
@@ -244,9 +269,13 @@ def flush
         if self.class.browser_security_headers.include?(property) && !data.key?('browserSecurityHeaders')
           data['browserSecurityHeaders'] = {}
         end
+        if self.class.smtp_server_properties.include?(property)
```
treydock, line 272: Same as other comment:

This should be self.class.smtp_server_properties.include?(property) && !data.key?('smtpServer') and then if this is true it adds the hash data['smtpServer'] = {} so that you are only adding the key if SMTP server settings exist. Then this current assignemtn gets moved to line 279.

treydock:
Looks good, just two minor corrections in the provider.  Also would be good if added acceptance tests.  See `spec/acceptance/2_realm_spec.rb`. Feel free to add SMTP settings similar to what you have deployed to verify functionality that gets tested inside docker.

mattock:
Thanks for the review! I will fix those two later this week. Then I need to sort out my development environment in order to be able to add and test the acceptance tests :smile:.

treydock:
I use RVM for Ruby and then you'll also need docker.  You might be able to use PDK to avoid needing RVM but I've never tested that.

lib/puppet/provider/keycloak_realm/kcadm.rb
mattock: This one has now been done as requested. Did not seem to cause any issues.

lib/puppet/provider/keycloak_realm/kcadm.rb
mattock: I implemented this as requested but (fortunately) noticed a glitch as I was writing acceptance tests. If done as you suggested changing the property values will actually destroy all the properties that do not change. Suppose I run this code to create a new realm:
```
keycloak_realm { 'foobar':
  ensure => 'present',
  smtp_server_host => 'server.example.org',
  smtp_server_port => '25',
}
```
Then I change  _smtp_server_host_  to _foobar.example.org_ what I end up with in the realm SMTP settings is this:

```
"smtpServer" : {
    "host" : "foobar.example.org"
}
```
So the "port" is gone from the realm.

I ran out of time today and could not debug this further, but the original/current version I have here does work.

Not sure if this affects the flush implementation, but when a new realm is created using kcadm an empty smtpServer hash is always present:

```
$ /opt/keycloak/bin/kcadm-wrapper.sh create realms -s realm=foobar -s enabled=true -o|grep smtpServer
Logging into http://localhost:8080/auth as user admin of realm master
  "smtpServer" : { },
```
Any suggestions would of course be welcome.

mattock:
The acceptance tests now work as expected. One test checks the create method and another one checks the flush implementation. The flush checks changes to three properties (a boolean, an integer and one with long property name).

lib/puppet/provider/keycloak_realm/kcadm.rb
treydock: The reason you lost items is because the line 275 condition only looks at properties if they are defined in `@property_flush` which gets populated only with changed things.  It might be possible to work around this issue by removing commented part on line 275 to also check `resource` which would allow you to make my suggested changes and not end up with some items missing if they don't change.  The acceptance tests should tell if you removing that comment breaks things or not.

I'll add some suggested changes

lib/puppet/provider/keycloak_realm/kcadm.rb
```diff
@@ -234,6 +258,7 @@ def initialize(value = {})
   def flush
     unless @property_flush.empty?
       data = {}
+      data['smtpServer'] = {}
```
treydock, line 261:
```suggestion
```

lib/puppet/provider/keycloak_realm/kcadm.rb
```diff
@@ -244,9 +269,13 @@ def flush
         if self.class.browser_security_headers.include?(property) && !data.key?('browserSecurityHeaders')
           data['browserSecurityHeaders'] = {}
         end
+        if self.class.smtp_server_properties.include?(property)
+          data['smtpServer'][camelize(property.to_s.gsub(%r{smtp_server_}, ''))] = resource[property] if resource[property]
+        end
```
treydock, line 274:
```suggestion
        if self.class.smtp_server_properties.include?(property) && !data.key?('smtpServer')
          data['smtpServer'] = {}
        end
```

lib/puppet/provider/keycloak_realm/kcadm.rb
```diff
@@ -244,9 +269,13 @@ def flush
...
         if @property_flush[property.to_sym] # || resource[property.to_sym]
```
treydock, line 275:
```suggestion
        if @property_flush[property.to_sym] || resource[property.to_sym]
```

lib/puppet/provider/keycloak_realm/kcadm.rb
```diff
@@ -244,9 +269,13 @@ def flush
...
           if self.class.browser_security_headers.include?(property)
             data['browserSecurityHeaders'][camelize(property)] = convert_property_value(resource[property.to_sym])
+          elsif self.class.smtp_server_properties.include?(property)
           else
```
treydock, line 279:
```suggestion
          elsif self.class.smtp_server_properties.include?(property)
            data['smtpServer'][camelize(property.to_s.gsub(%r{smtp_server_}, ''))] = resource[property]
          else
```

lib/puppet/type/keycloak_realm.rb
```diff
@@ -186,4 +186,69 @@
...
+  end
+
+  newproperty(:smtp_server_host) do
+    desc 'smtpServer host'
+  end
+
+  newproperty(:smtp_server_port) do
+    desc 'smtpServer port'
+    munge { |v| v.to_s }
+  end
```
treydock, line 218:
```suggestion
  newproperty(:smtp_server_port, parent: PuppetX::Keycloak::IntegerProperty) do
    desc 'smtpServer port'
  end
```

treydock: This is proper way wit this module to make something an integer.

mattock: Ok, thanks, will fix.

lib/puppet/provider/keycloak_realm/kcadm.rb
mattock: Thanks, I'll see if that breaks things and report back.

mattock:
@treydock may I inquire which commands you typically run to run all the validation and spec tests locally on your workstation? Although I was able to figure out some incantations for doing validation (rubocop et al) and to run Beaker acceptance tests I'm not sure if they're optimal.

I'd love to issue a PR which documents the test environment setup and how to run the tests. After all, I may have to write and run tests again later when I've forgotten all the details already :smile: . I generally just run "pdk validate" and "pdk test unit" in other projects and this is my first real encounter with beaker-rspec.

treydock:
I am fairly certain `pdk validate` and `pdk test unit` will still work on this module but the beaker actually spawns docker containers and applies the acceptance test code to a running system. I use a command like this:

```
BEAKER_set=centos-7 BEAKER_keycloak_full=true BEAKER_destroy=no bundle exec rake beaker
```

You can omit the `BEAKER_keycloak_full` if you don't want to test types, those take longer.  This PR does need to test types but you could limit scope and do this to just run realm tests:

```
BEAKER_set=centos-7 BEAKER_keycloak_full=true BEAKER_destroy=no bundle exec rspec spec/acceptance/2_realm_spec.rb
```

By doing `BEAKER_destroy=no` you are able to connect to the container after a failed test like so 

```
docker exec -it keycloak-el7 /bin/bash
```

The best place for such docs would be CONTRIBUTING.md I think, as README is more for usage and not development.  Would need to unmanage config file in .sync.yml to avoid PDK sync from modifying changes:

```yaml
CONTRIBUTING.md:
  unmanaged: true
```

At some point I would then copy changes to my pdk template so could re-sync with other modules.

mattock:
Ok, sounds good. I'll go through the test environment setup on a fresh VM so that I can properly document it.

mattock:
I made the suggested changes. Due to Rubocop warnings I had to remove one layer of ifs and use && instead. Functionality did not change.

All the keycloak_realm acceptance tests now pass:
```
--- snip ---
Finished in 5 minutes 20 seconds (files took 2 minutes 11.3 seconds to load)
12 examples, 0 failures
```
Is this perhaps good enough to be merged now :smile:?

treydock:
PR looks good but unit tests are failing.  I am having issues seeing why because of some issues with Travis-CI it seems.  I will merge and fix up any lingering issues in a separate pull request if there are any.  I was not seeing any failures when I ran tests locally.

PR #136, opened 2020-06-06
Fix lint warning and document PDK setup

mattock:


CONTRIBUTING.md
```diff
@@ -1,3 +1,16 @@
+# PDK setup
+
+PDK needs the bundler gem for each Ruby version tested in this module. For
+example:
+
+    $ sudo /opt/puppetlabs/pdk/private/ruby/2.4.9/bin/gem install bundler
+    $ sudo /opt/puppetlabs/pdk/private/ruby/2.5.7/bin/gem install bundler
```
treydock, line 7: I don't think these are necessary.  I think bundler will always be present, or should be. I've never had to run this step before.  Also I don't think it's necessary to test each version of Ruby with PDK, just test whatever is the default and that should work 99% of the time.  I think the step that is necessary is

```
pdk bundle install
```

treydock: It also might be the case that `pdk bundle install` happens automatically when you run the other pdk commands.

mattock: Oh, I did not even know about ```pdk bundle install```. That said, "pdk validate" etc. were failing for me on a "out of the box" PDK.

treydock: In what way was `pdk validate` failing? What was the error?  Normally PDK should require very little extra setup to get working with PDK enabled modules.

mattock: I'll open an issue on that. Other pdk commands such "pdk test unit" were also failing, even after bundler was installed. I'll make that a separate ticket.

PR #143, opened 2020-07-04
Update usage of deprecated function postgresql_password

Karlinde:
The function `postgresql_password` was recently deprecated in puppetlabs-postgresql: https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/REFERENCE.md#postgresql_password, causing a deprecation warning.

Suggested replacement: `postgresql::postgresql_password`


treydock:
The tests are failing because you need to update `.fixtures.yml` version to at `v6.4.0` which is when this function was introduced. Also need to update `metadata.json` to increase minimum version of dependency to `6.4.0`.

PR #146, opened 2020-08-02
Explicitly specifies what user to use with the admin generation script

ZloeSabo:
**Description**
This change explicitly specifies what user to use with the admin generation script,
so the keycloak would have no issues accessing generated keycloak-add-user.json.

**Background**
The add-user-keycloak generates keycloak-add-user.json belonging to a user that runs the script.
This means if puppet and keycloak services belong to different users,
without this change there was no guarantee that keycloak could access to this file after creation.


treydock:
@ZloeSabo Thanks, this will be released in v6.13.1

PR #148, opened 2020-08-07
Add support for required actions

ZloeSabo:
**Description**
Required actions are actions a user must perform during the authentication process. A user will not be able to complete the authentication process until these actions are complete. For instance, change a one-time password, accept T&C, etc.

**Note**
I deliberately did not add auto requires as it complicates the deconfiguration process. I.e. if I specify `ensure => absent` for both required_action and realm and the former auto requires the latter, then the realm would be deleted first and the puppet run would fail on deleting actions because they're specific to a realm and the realm no longer exists.


aba-rechsteiner:
+1

PR #150, opened 2020-08-12
add custom config lines

aba-rechsteiner:


treydock:
What does this add that `custom_config_content` does not do? Is it the ability to supply an array?  If the goal here is to supply an array of lines I'd suggest updating `custom_config_content` to support Array then do something like this here: https://github.com/treydock/puppet-module-keycloak/blob/master/manifests/config.pp#L71

```
if $keycloak::custom_config_content or $keycloak::custom_config_source {
  if $keycloak::custom_config_content =~ Array {
    $custom_config_content = join($keycloak::custom_config_content, "\n")
  } else {
    $custom_config_content = $keycloak::custom_config_content
  }

    concat::fragment { 'config.cli-custom':
...
          content => $custom_config_content,
} 

```

aba-rechsteiner:
ah sorry, i missed the concat::fragment. i thought it would replace the whole file

PR #153, opened 2020-08-21
Fixing wrong filename in module.xml for datasource oracle

zaeh:
The filename of the driver jar for datasource oracle is fixed to "oracle.jar" in  class keycloak::datasource::oracle.
This is not working with the fixed filename ojdbc8.jar in module.xml.


zaeh:
This is the way I need to do this at the moment, because of the bug.
The datasource "oracle" can not work at all without an own file/template for the module.xml given to the keycloak class as:
datasource_module_source => 'puppet:///<path of custom module.xml>'

As an alternative you could parameterize the value for the oracle.jar to make this working.

treydock:
Right, my suggestion was parameterize the value. The Oracle support was a community contribution and I don't want to break other people's code in case their jar name is somehow different. If you it a parameter that would solve the problem and avoid breaking changes.

treydock:
I am looking at the code, are you using `datasource_jar_source`?  If you are using that parameter then `oracle.jar` should exist.  I think the only reliable way is to ensure `datasource_jar_source` is set.

treydock:
Just FYI that this change would most definitely break anyone using `datasource_jar_source` because that parameter will enforce the name of the jar to be `oracle.jar`.

zaeh:
Exactly that ist the problem:
1. If you are using `datasource_jar_source` the name of the jar is fixed to `oracle.jar`
2. The `module.xml` that ist used contains the hardcoded filename `ojdbc8.jar`
3. JBoss will not start because it is trying to load `ojdbc8.jar` but the driver is named `oracle.jar`

zaeh:
Done.
What do you think?

manifests/datasource/oracle.pp
```diff
@@ -5,7 +5,7 @@
 class keycloak::datasource::oracle {
   assert_private()
 
-  $module_source = pick($keycloak::datasource_module_source, 'puppet:///modules/keycloak/database/oracle/module.xml')
+  $jar_filename = pick($keycloak::datasource_jar_filename, 'oracle.jar')
```
treydock, line 8:
```suggestion
  $jar_filename = pick($keycloak::datasource_jar_filename, 'ojdbc8.jar')
```

To maintain backwards compatibility the default should remain `ojdbc8.jar`.

zaeh: Hmm, if anybody uses an own module.xml (via parameter `datasource_module_source`) with the filename corrected to the formerly hardcoded `oracle.jar`, this will fail for him/her completely without changing the name via the new parameter `datasource_jar_filename`.
That's not backwards compatible at all, I guess.

treydock: I think the old behavior worked because the person that contributed the code had `ojdbc8.jar`, not sure how they had that name but that's the name that was used in original pull request.  I think the suggested change is sufficient to maintain backwards compatibility even if it technically might not work. So to use `oracle.jar` you should only need to set `datasource_jar_filename`.  Feel free to update README with Oracle example that uses the code you are using to make things work, that way it's easier for others to see.

treydock:
@zaeh Looks good. Would you mind adding an example of how you are using Oracle support to README.md?  There are already examples for MySQL and PostgreSQL so can just add one example after those.

treydock:
@zaeh, this is ready to merge, just needs README update with example of using Oracle similar to examples for MySQL and PostgreSQL.

zaeh:
Sorry, I was in holidays... Will add it in the next days! :-)

treydock:
@zaeh Wanted to check if had a chance to make those README updates for this pull request.

zaeh:
Sorry, that it took so long!
I was very busy and just forgot about it.
Hope you like the Oracle usage example in the Readme.

treydock:
This will be released with v6.21.0

PR #155, opened 2020-09-18
add keycloak-oidc providerid and other new parameters

aba-rechsteiner:


lib/puppet/type/keycloak_identity_provider.rb
```diff
@@ -258,7 +276,7 @@ def self.title_patterns
     if self[:realm].nil?
       raise Puppet::Error, 'realm is required'
     end
-    if self[:ensure].to_s == 'present' && self[:provider_id] == 'oidc'
+    if self[:ensure].to_s == 'present' && (self[:provider_id] == 'oidc' || self[:provider_id] == 'keycloak-oidc')
```
treydock, line 279: Might be shorter to say `['oidc', 'keycloak-oidc'].include?(self[:provider_id])` and if another provider_id is added it would be simpler to add new option without having extremely log line.

treydock:
@aba-rechsteiner Could you please add some acceptance tests to `spec/acceptance/8_identity_provider_spec.rb` and appropriate additions to unit tests under `spec/unit/puppet/type/keycloak_identity_provider_spec.rb`. If you need any help with that let me know.

treydock:
Also once changes are made please squash commits into single commit.

PR #163, opened 2020-10-05
Add more realm login related properties

anlambert:
That PR enables to set `resetPasswordAllowed` and `verifyEmail` options to a realm from Puppet.


treydock:
This will be released with v6.19.0

PR #164, opened 2020-10-05
Enable roles management at realm and client level

anlambert:
That PR enables to create/delete [roles](https://www.keycloak.org/docs/6.0/server_admin/#roles) at realm and client level.

Only simple (non composite) roles are considered.

Composite roles management could be added later by adding a new `keycloak_composite_role` type
setting the mapping between the new composite role and a list of other existing roles.

### DIFF
--- a/lib/puppet/provider/keycloak_client/kcadm.rb
+++ b/lib/puppet/provider/keycloak_client/kcadm.rb
@@ -37,6 +37,26 @@ Puppet::Type.type(:keycloak_client).provide(:kcadm, parent: Puppet::Provider::Ke
     end
   end
 
+  def self.get_client_roles(realm, client)
+    output = kcadm('get', "clients/#{client}/roles", realm)
+    Puppet.debug("Client #{client} in realm #{realm} roles: #{output}")
+    data = JSON.parse(output)
+    roles = []
+    data.each do |d|
+      # filter out 'uma_protection' client role created when
+      # authorization_services_enabled property is set to true
+      if !d['composite'] && d['name'] != 'uma_protection'
+        roles.push(d['name'])
+      end
+    end
+    Puppet.debug("Returned client roles: #{roles}")
+    roles
+  end
+
+  def get_client_roles(*args)
+    self.class.get_client_roles(*args)
+  end
+
   def self.instances
     clients = []
     realms.each do |realm|
@@ -72,6 +92,7 @@ Puppet::Type.type(:keycloak_client).provide(:kcadm, parent: Puppet::Provider::Ke
         client[:realm] = realm
         client[:name] = "#{client[:client_id]} on #{client[:realm]}"
         type_properties.each do |property|
+          next if [:roles].include?(property)
           camel_key = camelize(property)
           dot_key = property.to_s.tr('_', '.')
           key = property.to_s
@@ -96,6 +117,7 @@ Puppet::Type.type(:keycloak_client).provide(:kcadm, parent: Puppet::Provider::Ke
         end
         # The absence of a value should be 'absent'
         client[:login_theme] = 'absent' if client[:login_theme].nil?
+        client[:roles] = get_client_roles(realm, client[:id])
         clients << new(client)
       end
     end
@@ -162,7 +184,7 @@ Puppet::Type.type(:keycloak_client).provide(:kcadm, parent: Puppet::Provider::Ke
     data[:clientId] = resource[:client_id]
     data[:secret] = resource[:secret] if resource[:secret]
     type_properties.each do |property|
-      next if [:default_client_scopes, :optional_client_scopes].include?(property)
+      next if [:default_client_scopes, :optional_client_scopes, :roles].include?(property)
       next unless resource[property.to_sym]
       value = convert_property_value(resource[property.to_sym])
       next if value == 'absent' || value == :absent || value.nil?
@@ -243,6 +265,33 @@ Puppet::Type.type(:keycloak_client).provide(:kcadm, parent: Puppet::Provider::Ke
         raise Puppet::Error, "kcadm update clients/#{resource[:id]}/optional-client-scopes/#{scope_id}: #{e.message}"
       end
     end
+    role = nil
+    if resource[:roles]
+      roles = get_client_roles(resource[:realm], resource[:id])
+      remove_roles = roles - resource[:roles]
+      begin
+        remove_roles.each do |s|
+          role = s
+          kcadm('delete', "clients/#{resource[:id]}/roles/#{role}", resource[:realm])
+        end
+      rescue Puppet::ExecutionFailure => e
+        raise Puppet::Error, "kcadm delete realms/#{resource[:realm]}/clients/#{resource[:id]}/roles/#{role}: #{e.message}"
+      end
+      add_roles = resource[:roles] - roles
+      begin
+        add_roles.each do |s|
+          role = s
+          role_data = { 'description' => "${role_#{role}}", 'name' => role }
+          role_data_t = Tempfile.new('keycloak_client_role')
+          role_data_t.write(JSON.pretty_generate(role_data))
+          role_data_t.close
+          Puppet.debug(IO.read(role_data_t.path))
+          kcadm('create', "clients/#{resource[:id]}/roles", resource[:realm], role_data_t.path)
+        end
+      rescue Puppet::ExecutionFailure => e
+        raise Puppet::Error, "kcadm create realms/#{resource[:realm]}/clients/#{resource[:id]}/roles/#{role}: #{e.message}"
+      end
+    end
     @property_hash[:ensure] = :present
   end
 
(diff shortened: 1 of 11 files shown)


lib/puppet/provider/keycloak_client/kcadm.rb
```diff
@@ -221,6 +242,19 @@ def create
         raise Puppet::Error, "kcadm delete clients/#{resource[:id]}/optional-client-scopes/#{scope_id}: #{e.message}"
       end
     end
+    role = nil
+    if resource[:roles]
+      roles = get_client_roles(resource[:realm], resource[:id])
+      remove_roles = roles - resource[:roles]
```
treydock, line 248: You shouldn't need to query roles again at this point, they will be stored in `@property_hash[:roles]` based on what was collected in `self.instances`.

Could make this:
```
remove_roles = @property_hash[:roles] - resource[:roles]
```

lib/puppet/provider/keycloak_client/kcadm.rb
```diff
@@ -243,6 +277,23 @@ def create
         raise Puppet::Error, "kcadm update clients/#{resource[:id]}/optional-client-scopes/#{scope_id}: #{e.message}"
       end
     end
+    if resource[:roles]
+      roles = get_client_roles(resource[:realm], resource[:id])
+      add_roles = resource[:roles] - roles
```
treydock, line 282:
```suggestion
      add_roles = resource[:roles] - @property_hash[:roles]
```

lib/puppet/provider/keycloak_client/kcadm.rb
```diff
@@ -343,6 +394,19 @@ def flush
           raise Puppet::Error, "kcadm delete clients/#{id}/optional-client-scopes/#{scope_id}: #{e.message}"
         end
       end
+      role = nil
+      if @property_flush[:roles]
+        roles = get_client_roles(resource[:realm], resource[:id])
+        remove_roles = roles - @property_flush[:roles]
```
treydock, line 400:
```suggestion
        remove_roles = @property_hash[:roles] - @property_flush[:roles]
```

lib/puppet/provider/keycloak_client/kcadm.rb
```diff
@@ -365,6 +429,23 @@ def flush
           raise Puppet::Error, "kcadm update clients/#{id}/optional-client-scopes/#{scope_id}: #{e.message}"
         end
       end
+      if @property_flush[:roles]
+        roles = get_client_roles(resource[:realm], resource[:id])
+        add_roles = @property_flush[:roles] - roles
```
treydock, line 434:
```suggestion
        add_roles = @property_flush[:roles] - @property_hash[:roles]
```

lib/puppet/provider/keycloak_realm/kcadm.rb
```diff
@@ -202,6 +223,19 @@ def create
         raise Puppet::Error, "kcadm delete realms/#{resource[:name]}/default-optional-client-scopes/#{scope_id}: #{e.message}"
       end
     end
+    role = nil
+    if resource[:roles]
+      roles = get_realm_roles(resource[:name])
+      remove_roles = roles - resource[:roles]
```
treydock, line 229:
```suggestion
      remove_roles = @property_hash[:roles] - resource[:roles]
```

lib/puppet/provider/keycloak_realm/kcadm.rb
```diff
@@ -226,6 +260,23 @@ def create
         raise Puppet::Error, "kcadm update realms/#{resource[:name]}/default-optional-client-scopes/#{scope_id}: #{e.message}"
       end
     end
+    if resource[:roles]
+      roles = get_realm_roles(resource[:name])
+      add_roles = resource[:roles] - roles
```
treydock, line 265:
```suggestion
      add_roles = resource[:roles] - @property_hash[:roles]
```

lib/puppet/provider/keycloak_realm/kcadm.rb
```diff
@@ -341,6 +392,19 @@ def flush
           raise Puppet::Error, "kcadm delete realms/#{resource[:name]}/default-optional-client-scopes/#{scope_id}: #{e.message}"
         end
       end
+      role = nil
+      if @property_flush[:roles]
+        roles = get_realm_roles(resource[:name])
+        remove_roles = roles - @property_flush[:roles]
```
treydock, line 398:
```suggestion
        remove_roles = @property_hash[:roles] - @property_flush[:roles]
```

lib/puppet/provider/keycloak_realm/kcadm.rb
```diff
@@ -365,6 +429,23 @@ def flush
           raise Puppet::Error, "kcadm update realms/#{resource[:name]}/default-optional-client-scopes/#{scope_id}: #{e.message}"
         end
       end
+      if @property_flush[:roles]
+        roles = get_realm_roles(resource[:name])
+        add_roles = @property_flush[:roles] - roles
```
treydock, line 434:
```suggestion
        add_roles = @property_flush[:roles] - @property_hash[:roles]
```

treydock:
At least one of the failures in acceptance tests is related to this PR, another might be unrelated but not sure.  Looks like the client role is not being idempotent.

lib/puppet/provider/keycloak_client/kcadm.rb
```diff
@@ -221,6 +242,19 @@ def create
         raise Puppet::Error, "kcadm delete clients/#{resource[:id]}/optional-client-scopes/#{scope_id}: #{e.message}"
       end
     end
+    role = nil
+    if resource[:roles]
+      roles = @property_hash[:roles] ||= get_client_roles(resource[:realm], resource[:id])
```
treydock, line 247: I think my previous recommendation was wrong.  `@property_hash` won't be set during `create` because `create` is called when something does not exist.  So your previous approach was correct, only do this:

```
role = get_client_roles(resource[:realm], resource[:id])
```

lib/puppet/provider/keycloak_client/kcadm.rb
```diff
@@ -243,6 +277,23 @@ def create
         raise Puppet::Error, "kcadm update clients/#{resource[:id]}/optional-client-scopes/#{scope_id}: #{e.message}"
       end
     end
+    if resource[:roles]
+      roles = @property_hash[:roles] ||= get_client_roles(resource[:realm], resource[:id])
```
treydock, line 281: I think my previous recommendation was wrong.  `@property_hash` won't be set during `create` because `create` is called when something does not exist.  So your previous approach was correct, only do this:

```
role = get_client_roles(resource[:realm], resource[:id])
```

Also any reason not to move this logic under same `if resource[:roles]` ? That would make the code a little less duplicated.

lib/puppet/provider/keycloak_realm/kcadm.rb
```diff
@@ -202,6 +223,19 @@ def create
         raise Puppet::Error, "kcadm delete realms/#{resource[:name]}/default-optional-client-scopes/#{scope_id}: #{e.message}"
       end
     end
+    role = nil
+    if resource[:roles]
+      roles = @property_hash[:roles] ||= get_realm_roles(resource[:name])
```
treydock, line 228: Per other comment , just make this :

```
role = get_realm_roles(resource[:name])
```

lib/puppet/provider/keycloak_realm/kcadm.rb
```diff
@@ -226,6 +260,23 @@ def create
         raise Puppet::Error, "kcadm update realms/#{resource[:name]}/default-optional-client-scopes/#{scope_id}: #{e.message}"
       end
     end
+    if resource[:roles]
+      roles = @property_hash[:roles] ||= get_realm_roles(resource[:name])
```
treydock, line 264: Per other comment , just make this :

```
role = get_realm_roles(resource[:name])
```

Also move this logic under the other `if resource[:roles]` so only doing the `get_realm_roles` query once.

lib/puppet/provider/keycloak_realm/kcadm.rb
```diff
@@ -365,6 +428,22 @@ def flush
           raise Puppet::Error, "kcadm update realms/#{resource[:name]}/default-optional-client-scopes/#{scope_id}: #{e.message}"
         end
       end
+      if @property_flush[:roles]
+        add_roles = @property_flush[:roles] - @property_hash[:roles]
```
treydock, line 432: Move this logic up under same `if @property_flush[:roles]` so that the logic is all together.

anlambert:
@treydock , thanks for the review ! My last update should address all of your comments. There is still an acceptance that might faill, I did not yet understand why but this is clearly related to my last changes as the first version of that PR passed all travis tests.

anlambert:
> There is still an acceptance that might faill, I did not yet understand why but this is clearly related to my last changes as the first version of that PR passed all travis tests.

I found the issue. When the `authorization_services_enabled` property is set to `true` for a client, Keycloak will create the `uma_protection` role when updating a client and thus the acceptance test fails due to that unexpected extra role. I have updated the PR by filtering out that special role when getting client roles.

treydock:
This will be released with v6.19.0

PR #168, opened 2020-11-16
Support realm remember_me parameters

mattock:
This adds suppor for ssoSessionIdleTimeoutRememberMe and
ssoSessionMaxLifespanRememberMe.


treydock:
This will be released with v6.22.0

PR #170, opened 2020-12-04
Add convenience define for setting up FreeIPA LDAP mappers

mattock:
The mappings are well-known, so there's no reason for all FreeIPA users to set
them up separately.

### DIFF
--- a/manifests/freeipa_ldap_mappers.pp
+++ b/manifests/freeipa_ldap_mappers.pp
@@ -3,43 +3,40 @@
 #
 # @example
 #   keycloak::freeipa_ldap_mappers { 'ipa.example.org':
-#     ldap_provider_id => '6bd0f68d-5618-a74d-bcab-089702f1bc80',
 #     realm            => 'EXAMPLE.ORG',
 #     groups_dn        => 'cn=groups,cn=accounts,dc=example,dc=org',
 #     roles_dn         => 'cn=groups,cn=accounts,dc=example,dc=org'
 #   }
 #
-# @param parent_name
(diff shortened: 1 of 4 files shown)


manifests/freeipa_ldap_mappers.pp
```diff
@@ -0,0 +1,120 @@
...
+# @param ldap_provider_id
+#   Identifier for the LDAP provider to add this mapper to
+#
+define keycloak::freeipa_ldap_mappers
+(
+  String $realm,
+  String $groups_dn,
+  String $roles_dn,
+  String $parent_name = $title,
+  String $ldap_provider_id = $title,
```
treydock, line 30: I think you'll have to make `ldap_provider_id` be `Optional[String] $ldap_provider_id = undef` and then in the manifest do something like this:

```puppet
$_ldap_provider_id = pick($ldap_provider_id, "${title}-${realm}")
```

The default is here: https://github.com/treydock/puppet-module-keycloak/blob/master/lib/puppet/type/keycloak_ldap_user_provider.rb#L33-L38

treydock:
Be good to have unit tests and acceptance tests and mention on README.  Similar to #135 you can just add the acceptance tests to https://github.com/treydock/puppet-module-keycloak/blob/master/spec/acceptance/3_ldap_spec.rb to simplify things.

mattock:
Ok, I'll check those out next week. We want to get all our stuff upstream before we end up with a merge hell :smile:.

treydock:
Released with v6.23.0

mattock:
Thanks @treydock!

PR #172, opened 2020-12-08
Fix puppet-lint warning

mattock:
Found by "pdk validate".


treydock:
Released with v6.23.0

PR #180, opened 2021-01-29
Domain mode support

mattock:
This PR adds support for setting up Keycloak domain mode clusters.

Some notes:

* The Vagrant environment for developing domain mode code is much more complex than the Vagrant code used by standalone Keycloaks. For this reason it is located in a separate subdirectory. It could be renamed to something else like "domain-mode-vagrant". Or we could merge all the Vagrant code together. We can also split out the Vagrant environment from this PR, if needed.
* Acceptance tests only cover the domain master because a domain slave will refuse to start if master is unreachable. The only way to fix this would be to do multi-machine Beaker tests. The documentation on multi-machine Beaker seems very sketchy, especially when it comes to the Docker backend. If you have any pointers let me know.


treydock:
The acceptance test approach is rather involved. It requires a different YAML file for defining the containers, here's an example: https://github.com/sensu/sensu-puppet/blob/master/spec/acceptance/nodesets/centos-7.yml. For this module you'd name it like `centos-7-cluster.yml` or `centos-8-cluster.yml` and define 2 containers in that YAML, they have to have `agent` role and then one has role like `node1` and other role is `node2`, or some other unique name between each.

Then need to add a new mode like this: https://github.com/treydock/puppet-module-keycloak/blob/master/spec/spec_helper_acceptance_setup.rb#L10. Maybe named `cluster` or something so like `BEAKER_keycloak_cluster` environment variable saved to `keycloak_cluster`.  Then a new acceptance file with something like this:

```
describe 'keycloak cluster', if: RSpec.configuration.keycloak_cluster do
...
end
```

Then in files like here need `unless: RSpec.configuration.keycloak_cluster`: https://github.com/treydock/puppet-module-keycloak/blob/master/spec/acceptance/1_class_spec.rb#L3

Then you have to add the tests to Github Actions so if you have PDK you can add something like this here: https://github.com/treydock/puppet-module-keycloak/blob/master/.sync.yml#L28

```
    - set: centos-7-cluster
      puppet: puppet5
      keycloak_version: 8.0.1
      keycloak_cluster: 'yes'
    - set: centos-7-cluster
      puppet: puppet5
      keycloak_version: 12.0.1
      keycloak_cluster: 'yes'
    - set: centos-7-cluster
      puppet: puppet6
      keycloak_version: 8.0.1
      keycloak_cluster: 'yes'
    - set: centos-7-cluster
      puppet: puppet6
      keycloak_version: 12.0.1
      keycloak_cluster: 'yes'
```

Then need something like this: https://github.com/treydock/puppet-module-keycloak/blob/master/.sync.yml#L27 but for `keycloak_cluster`.  Then can run `pdk update` which might change more things so might be better to just make changes to this file: https://github.com/treydock/puppet-module-keycloak/blob/master/.github/workflows/ci.yaml#L52

treydock:
Also I would recommend where possible if a resource exists in both `keycloak::config::standalone` and `keycloak::config::domain` then that would be in like `keycloak::config::common` or just `keycloak::config` and then just use some logic to change parameter values for those resources.  This is to avoid too much duplication.  I think there is enough common resources between them that should be able to keep everything in `keycloak::config` and just change parameters based on whether domain or standalone.

.fixtures-latest.yml
```diff
@@ -17,6 +17,12 @@ fixtures:
     yumrepo_core:
       repo: https://github.com/puppetlabs/puppetlabs-yumrepo_core.git
       puppet_version: ">= 6.0.0"
+    augeas_core:
+      repo: https://github.com/puppetlabs/puppetlabs-augeas_core.git
+      puppet_version: ">= 6.0.0"
+    yumrepo_core:
```
treydock, line 23: yumrepo_core is already in this file, so this addition is a duplicate of whats a few lines higher.

mattock: Ok, thanks for the detailed pointers to multi-container tests. I'll look into it. The other fixes you proposed also make sense, will look into them as well.

treydock:
Having looked more I think a pattern like this makes more sense inside `keycloak::config` :

```
if $::keycloak::operating_mode != 'domain' {
 # standalone and cluster things if appropriate
} else {
 # domain resources
}
```

I think in some cases the condition will be just variables like the paths used to setup admin user or to run jboss-cli.  Then other cases the condition would contain entire resources like which directories to create and for domain just the augeas resources.  There is just too much duplication with doing 2 distinct classes.  I think it's okay to have multiple places where the condition above is used in `keycloak::config` if it makes sense to keep some things logically separate.

.gitignore
```diff
@@ -25,3 +25,5 @@
 .project
 .envrc
 /inventory.yaml
+/vagrant/.vagrant/
+/vagrant/ubuntu-bionic-18.04-cloudimg-console.log
```
treydock, line 29: This can just be `/vagrant/*.log` I think

mattock:
Quick update: the multi-machine Beaker tests now work locally. I did a two-way replication check as well as the usual "is service up and enabled" type checks. Thanks for the excellent pointers @treydock, it really helped get this done fast!

I'll do the rest of the proposed changes first in our own GitHub fork and then update this PR. Hopefully that happens at some point tomorrow.

mattock:
@treydock I believe all the feedback you gave has been addressed.

manifests/config.pp
```diff
@@ -21,37 +21,47 @@
...
+  if $::keycloak::operating_mode != 'domain' {
+    $_add_user_keycloak_args = "--user ${keycloak::admin_user} --password ${keycloak::admin_user_password} --realm master"
+    $_subdir = 'standalone'
+    $_config_cli_content = template('keycloak/config.cli.erb')
+    $_java_opts_path = "${keycloak::install_base}/bin/standalone.conf"
+  } else {
+    $_server_conf_dir = "${keycloak::install_base}/domain/servers/${keycloak::server_name}/configuration"
+    $_add_user_keycloak_args = "--user ${keycloak::admin_user} --password ${keycloak::admin_user_password} --realm master --sc ${_server_conf_dir}/" # lint:ignore:140chars
+    $_subdir = 'domain'
+    $_config_cli_content = template('keycloak/config-domain.cli.erb')
```
treydock, line 43: Because this is an ERB template, it might make sense to stick with a single template for all modes and then just have something like this for the places that are specific to something like domain mode:

```
<%- if scope['keycloak::operating_mode'] == 'domain' -%>
// domain specific changes here
<%- end -%>
```

This will avoid a lot of duplication since the two templates do have a lot of overlap.

manifests/config.pp
```diff
@@ -64,7 +74,7 @@
 
   concat::fragment { 'config.cli-keycloak':
     target  => "${keycloak::install_base}/config.cli",
-    content => template('keycloak/config.cli.erb'),
```
treydock, line 67: Undo this once using single template

manifests/config.pp
mattock: :+1:

templates/config.cli.erb
```diff
@@ -60,7 +98,27 @@ catch
...
+/profile=auth-server-clustered/subsystem=datasources/data-source=KeycloakDS:write-attribute(name=flush-strategy, value=IdleConnections)
+try
+/profile=auth-server-clustered/subsystem=datasources/jdbc-driver=postgresql:add(driver-module-name=org.postgresql,driver-name=postgresql,driver-xa-datasource-class-name=org.postgresql.xa.PGXADataSource)
+catch
+/profile=auth-server-clustered/subsystem=datasources/jdbc-driver=postgresql:remove
+/profile=auth-server-clustered/subsystem=datasources/jdbc-driver=postgresql:add(driver-module-name=org.postgresql,driver-name=postgresql,driver-xa-datasource-class-name=org.postgresql.xa.PGXADataSource)
+end-try
 <%- end -%>
+
+<%- if scope['keycloak::operating_mode'] != 'domain' %>
```
treydock, line 121:
```suggestion
<%- if scope['keycloak::operating_mode'] != 'domain' -%>
```

templates/config.cli.erb
```diff
@@ -2,8 +2,30 @@
...
+if (outcome == success) of /profile=load-balancer:read-resource
+/profile=load-balancer:remove
+end-if
+if (outcome == success) of /socket-binding-group=load-balancer-sockets:read-resource
+/socket-binding-group=load-balancer-sockets:remove
+end-if
+if (outcome != success) of /socket-binding-group=ha-sockets/socket-binding=proxy-https:read-resource
+/socket-binding-group=ha-sockets/socket-binding=proxy-https:add(port=443)
+end-if
+<% end -%>
```
treydock, line 25:
```suggestion
<%- end -%>
```

templates/config.cli.erb
```diff
@@ -2,8 +2,30 @@
 embed-server
 <% elsif scope['keycloak::operating_mode'] == 'clustered'-%>
 embed-server --server-config=standalone-ha.xml
+<% else -%>
+embed-host-controller
 <% end -%>
-<%- if scope['keycloak::proxy_https'] -%>
+
+<% if scope['keycloak::operating_mode'] == 'domain' -%>
```
treydock, line 9:
```suggestion
<%- if scope['keycloak::operating_mode'] == 'domain' -%>
```

templates/config.cli.erb
```diff
@@ -74,7 +132,27 @@ end-if
 if (outcome == success) of /subsystem=keycloak-server/spi=truststore:read-resource
 /subsystem=keycloak-server/spi=truststore/:remove
 end-if
-<%- end -%>
+<%- end -%><%- # end keystore -%>
+<%- end -%><%- # end is not domain -%>
+
+<%- if scope['keycloak::operating_mode'] == 'domain' %>
```
treydock, line 138:
```suggestion
<%- if scope['keycloak::operating_mode'] == 'domain' -%>
```

treydock:
Made a few suggested changes, mostly to help ensure that unnecessary white lines are not added when this change is merged.  Ideally sites not using domain mode would have no changes to their config file and thus not force a restart of Keycloak service.

Also the failures are unrelated to your changes. I have other modules that are failing in a similar way and I've not yet been able to identify the issue.

kibahop:
> Made a few suggested changes, mostly to help ensure that unnecessary white lines are not added when this change is merged. Ideally sites not using domain mode would have no changes to their config file and thus not force a restart of Keycloak service.
> 
> Also the failures are unrelated to your changes. I have other modules that are failing in a similar way and I've not yet been able to identify the issue.

Yes. I hope I have fixed those now. The template is now unfortunately pretty horrible and unreadable, and could have a better control flow and DRY.

I identified what causes the issue with at least with puppet version 6.21.0. It is the forcelocal attribute to the user resource. With puppet version 5.5.22 this does not happen.

treydock:
I opened a ticket earlier today and it appears to have pull requests merged to fix the bug: https://tickets.puppetlabs.com/browse/PUP-10896.

templates/config-domain.cli.erb
```diff
@@ -0,0 +1,165 @@
+embed-host-controller --std-out=echo
```
treydock, line 1: This file should be removed as no longer needed.

templates/config.cli.erb
```diff
@@ -60,7 +98,27 @@ catch
...
+/profile=auth-server-clustered/subsystem=datasources/data-source=KeycloakDS:write-attribute(name=flush-strategy, value=IdleConnections)
+try
+/profile=auth-server-clustered/subsystem=datasources/jdbc-driver=postgresql:add(driver-module-name=org.postgresql,driver-name=postgresql,driver-xa-datasource-class-name=org.postgresql.xa.PGXADataSource)
+catch
+/profile=auth-server-clustered/subsystem=datasources/jdbc-driver=postgresql:remove
+/profile=auth-server-clustered/subsystem=datasources/jdbc-driver=postgresql:add(driver-module-name=org.postgresql,driver-name=postgresql,driver-xa-datasource-class-name=org.postgresql.xa.PGXADataSource)
+end-try
 <%- end -%>
+
+<%- if scope['keycloak::operating_mode'] != 'domain' -%>
```
treydock, line 121: Could change this to something like

```
<% if scope['keycloak::truststore'] && if scope['keycloak::operating_mode'] != 'domain' -%>

<% else -%>

<% end -%>
```

treydock: There is another condition for if is domain...I wonder if instead of multiple blocks for different lines the condition is actually

```
if domain
  profile = '/profile=auth-server-clustered'
else
  profile = ''
end
```

Something like that since that seems to be only difference at first glance but maybe there are other things different.

kibahop:
> I opened a ticket earlier today and it appears to have pull requests merged to fix the bug: https://tickets.puppetlabs.com/browse/PUP-10896.

That is great. Well done.

metadata.json
```diff
@@ -28,6 +28,14 @@
       "name": "puppetlabs/java_ks",
       "version_requirement": ">= 1.0.0 <4.0.0"
     },
+    {
+      "name": "puppetlabs/augeas_core",
+      "version_requirement": ">= 1.0.0 <4.0.0"
+    },
+    {
+      "name": "puppetlabs/yumrepo_core",
+      "version_requirement": ">= 1.0.0 <4.0.0"
```
treydock, line 37: This should likely be `>= 1.0.0 <2.0.0`.  I'm not certain this is even needed though as it appears like fresh install of `puppet-agent` pulls in a copy of this module.

spec/acceptance/1_domain_mode_cluster_spec.rb
```diff
@@ -0,0 +1,136 @@
+require 'spec_helper_acceptance'
+
+# This check needs to be here or Beaker will try to run find_only_one on
+# non-domain-mode tests and fail miserably.
+if RSpec.configuration.keycloak_domain_mode_cluster
+  describe 'keycloak domain mode cluster' do
```
treydock, line 6: This should be

```
describe 'keycloak domain mode cluster' if: RSpec.configuration.keycloak_domain_mode_cluster
```

One fewer level of spacing for everything in this file.

manifests/config.pp
```diff
@@ -105,11 +114,163 @@
   } else {
     $_java_opts = $java_opts
   }
-  file_line { 'standalone.conf-JAVA_OPTS':
+  file_line { 'JAVA_OPTS':
```
treydock, line 117: This filename needs to somehow be specific to Keycloak to avoid potential conflicts with other modules or other Puppet code.  So maybe change to `keycloak-JAVA_OPTS`

treydock:
Added a few more things to change, acceptance tests should now be possible to pass if things actually work, the bug with Puppet has been fixed and new released pushed to repos.

metadata.json
kibahop: > Added a few more things to change, acceptance tests should now be possible to pass if things actually work, the bug with Puppet has been fixed and new released pushed to repos.

What you requested have now been done, I believe.

kibahop:
> Added a few more things to change, acceptance tests should now be possible to pass if things actually work, the bug with Puppet has been fixed and new released pushed to repos.

What you requested have now been done, I believe.

spec/acceptance/1_domain_mode_cluster_spec.rb
mattock: I had to revert the change you proposed here. As mentioned in the comments the if statement needs to be where it is - otherwise the find_only_one functions gets triggered even on non-domain-mode nodesets. I don't fully understand what causes that behavior.

mattock:
The .fixtures-latest.yml test breakages were caused by camptocamp-systemd dropping support for Puppet 5:
* https://github.com/camptocamp/puppet-systemd/commit/97dd16fa32886b5b0f77a6f38a4953d4c1511c0e

I locked the systemd module version to the commit preceding that commit. I hope that approach is ok for you, at least as a temporary measure. Imho fixing that more cleanly would be outside of the scope of this PR.

mattock:
@treydock is this good to merge? Or still something to fix?

treydock:
Testing a change in #183 using some tricks I've used in another module I had forgotten about to fix some of the acceptance testing conditions. Also updated .sync.yml to include changes to tests so they persist when I run `pdk update`.  Once #183 is passing and merged I'll get this released in a new release

treydock:
This will be released with v6.26.0 once the Github Actions complete.

mattock:
Excellent, thanks a bunch @treydock! We are very happy to see this mega-PR finally got merged :smile:.

PR #189, opened 2021-03-15
FreeIPA/LDAP provider related regression fixes

mattock:
This PR fixes a couple of issues encountered when testing the latest code against FreeIPA:
* The default value defined for realm roles made the keycloak_realm provider (want to) overwrite realm roles that came from FreeIPA (or any other LDAP server, for that matter). Workarounds like adding "roles => undef" or "roles => false" did not help. Simply removing the default value for "roles" the keycloak_realm type fixes this
* FreeIPA provider now supports full_sync_period and changed_sync_period parameter. Previously this was not necessary/possible as support for it was lacking from keycloak_ldap_user_provider.

### DIFF
--- a/lib/puppet/type/keycloak_realm.rb
+++ b/lib/puppet/type/keycloak_realm.rb
@@ -282,8 +282,21 @@ Manage Keycloak realms
     newvalues(:true, :false)
   end
 
+  newparam(:manage_roles, boolean: true) do
+    desc 'Manage realm roles'
+    newvalues(:true, :false)
+    defaultto(:true)
+  end
+
   newproperty(:roles, array_matching: :all, parent: PuppetX::Keycloak::ArrayProperty) do
     desc 'roles'
     defaultto ['offline_access', 'uma_authorization']
+
+    def insync?(is)
+      if resource[:manage_roles].to_s == 'false'
+        return true
+      end
+      super(is)
+    end
   end
 end
--- a/lib/puppet/provider/keycloak_realm/kcadm.rb
+++ b/lib/puppet/provider/keycloak_realm/kcadm.rb
@@ -247,7 +247,7 @@ Puppet::Type.type(:keycloak_realm).provide(:kcadm, parent: Puppet::Provider::Key
       end
     end
     role = nil
-    if resource[:roles]
+    if resource[:roles] && resource[:manage_roles].to_s == 'true'
       roles = get_realm_roles(resource[:name])
       remove_roles = roles - resource[:roles]
       begin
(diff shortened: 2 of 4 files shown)


lib/puppet/type/keycloak_realm.rb
```diff
@@ -284,6 +284,5 @@ def should_to_s(_newvalue)
 
   newproperty(:roles, array_matching: :all, parent: PuppetX::Keycloak::ArrayProperty) do
     desc 'roles'
-    defaultto ['offline_access', 'uma_authorization']
```
treydock, line 287: This default was something that is default when adding realms with Keycloak web UI IIRC so not sure it's good to change this.  I think a better solution would be to update provider to skip this.

Try this here: https://github.com/treydock/puppet-module-keycloak/blob/master/lib/puppet/provider/keycloak_realm/kcadm.rb#L162 and here: https://github.com/treydock/puppet-module-keycloak/blob/master/lib/puppet/provider/keycloak_realm/kcadm.rb#L321

```
if property.to_s == 'roles' && resource[:roles] == [false]
  next
end
```

The above idea would omit `roles` from the JSON uploaded to Keycloak which is sort of the same as removing a default and not defining a value.  Then you would say `roles => false`.  I think this works because this is an array property and so any single value gets converted to an array.

If this approach ends up working, I would recommend updating `desc` for the property to mention something about `false` being a way to not set roles.

mattock: Ok, I'll try that.

mattock: Did not work as planned. The roles will still get overwritten. I won't have time to look into this today, but if you have any suggestions let me know and I'll work on them tomorrow.

treydock: Upon further inspection, I see what is happening here.  I forgot that `roles` is not a property that sets something for realm resource, it adds / removes roles like here: https://github.com/treydock/puppet-module-keycloak/blob/master/lib/puppet/provider/keycloak_realm/kcadm.rb#L250

I think the right solution is a new parameter like this:

```ruby
newparam(:manage_roles, boolean: true) do
  desc 'Manage realm roles'
  newvalues(:true, :false)
  defaultto(:true)
end
```

Then in provider you do something like this:

```
if resource[:roles] && resource[:manage_roles].to_s == 'true'
```

Would have to do the above in `create` and something like this in `flush`:

```
if @property_flush[:roles] && resource[:manage_roles].to_s == 'true'
```

The only issue you may run into with this is now `roles` will think it needs to change when unmanaged so try this:

```ruby
  newproperty(:roles, array_matching: :all, parent: PuppetX::Keycloak::ArrayProperty) do
    desc 'roles'
    defaultto ['offline_access', 'uma_authorization']

    def insync?(is)
      if @resource[:manage_roles].to_s == 'true'
        return true
      end
      super(is)
    end 
  end
```

treydock:
@mattock Thanks for the pull request!  This is being released as v7.1.0.

PR #190, opened 2021-03-18
Add support for logging to syslog

kibahop:


manifests/init.pp
```diff
@@ -245,6 +245,22 @@
 #   IP address of the master in domain mode
 # @param server_name
 #   Server name in domain mode. Defaults to hostname.
+# @param syslog
+#   Enable syslog. Default false.
+# @param syslog_app_name
+#  Syslog app name. Default 'keycloak'.
+# @param syslog_facility
+#  Syslog facility. Default 'user-level'.
+# @syslog_hostname
```
treydock, line 254: This and other parameters need to be prefixed with `@param `

manifests/config.pp
```diff
@@ -173,6 +173,24 @@
     }
   }
 
+  if $keycloak::syslog {
```
treydock, line 176: Remove this condition, always manage this file.

templates/config.cli/12-syslog.epp
```diff
@@ -0,0 +1,23 @@
...
+if (outcome != success) of <%= $prefix -%>/subsystem=logging/syslog-handler=SYSLOG:read-resource
+<%= $prefix -%>/subsystem=logging/syslog-handler=SYSLOG:add
+<%= $prefix -%>/subsystem=logging/syslog-handler=SYSLOG/:write-attribute(name=app-name, value=<%= $syslog_app_name %>)
+<%= $prefix -%>/subsystem=logging/syslog-handler=SYSLOG/:write-attribute(name=facility, value=<%= $syslog_facility %>)
+<%= $prefix -%>/subsystem=logging/syslog-handler=SYSLOG/:write-attribute(name=hostname, value=<%= $syslog_hostname %>)
+<%= $prefix -%>/subsystem=logging/syslog-handler=SYSLOG/:write-attribute(name=level, value=<%= $syslog_level %>)
+<%= $prefix -%>/subsystem=logging/syslog-handler=SYSLOG/:write-attribute(name=port, value=<%= $syslog_port %>)
+<%= $prefix -%>/subsystem=logging/syslog-handler=SYSLOG/:write-attribute(name=server-address, value=<%= $syslog_server_address %>)
+<%= $prefix -%>/subsystem=logging/syslog-handler=SYSLOG/:write-attribute(name=syslog-format, value=<%= $syslog_format %>)
+end-if
```
treydock, line 22: Should make it so that if `$syslog` is `false` the syslog handler is removed.  Maybe this:

```
<%- } else { -%>
if (outcome == success) of <%= $prefix -%>/subsystem=logging/syslog-handler=SYSLOG:read-resource
<%= $prefix -%>/subsystem=logging/syslog-handler=SYSLOG:remove
end-if
```

templates/config.cli/12-syslog.epp
```diff
@@ -0,0 +1,23 @@
...
+String $syslog_facility,
+String $syslog_hostname,
+String $syslog_level,
+String $syslog_port,
+Stdlib::Host $syslog_server_address,
+String $syslog_format
+| -%>
+<%- if $syslog { -%>
+if (outcome != success) of <%= $prefix -%>/subsystem=logging/syslog-handler=SYSLOG:read-resource
+<%= $prefix -%>/subsystem=logging/syslog-handler=SYSLOG:add
```
treydock, line 14: The `outcome` condition should end after the `add`. If you move `end-if` to after this line then you ensure that changes to syslog options will actually change the file even if SYSLOG handler is already set.  I am fairly certain that `write-attribute` can be run over and over and not break if already exists.

manifests/init.pp
```diff
@@ -344,6 +360,14 @@
   Optional[Variant[String, Array]] $custom_config_source = undef,
   Optional[Stdlib::Host] $master_address = undef,
   String $server_name = $facts['hostname'],
+  Boolean $syslog = false,
+  String $syslog_app_name = 'keycloak',
+  String $syslog_facility = 'user-level',
+  String $syslog_hostname = $facts['fqdn'],
+  String $syslog_level = 'INFO',
+  String $syslog_port = '514',
```
treydock, line 368: Make this an integer value and of type `Stdlib::Port`

manifests/init.pp
```diff
@@ -344,6 +360,14 @@
   Optional[Variant[String, Array]] $custom_config_source = undef,
   Optional[Stdlib::Host] $master_address = undef,
   String $server_name = $facts['hostname'],
+  Boolean $syslog = false,
+  String $syslog_app_name = 'keycloak',
+  String $syslog_facility = 'user-level',
+  String $syslog_hostname = $facts['fqdn'],
```
treydock, line 366: This should be of type `Stdlib::Host`

treydock:
I would recommend adding an acceptance test so that in this context, https://github.com/treydock/puppet-module-keycloak/blob/master/spec/acceptance/1_class_spec.rb#L131 , change to set `syslog => true` and then maybe add another context after that one where the class definition is this:

```
      include mysql::server
      class { 'keycloak':
        datasource_driver => 'mysql',
      }
```

And call it like "reset back to defaults" or something.

So far looks like only remaining issues are data types used for `syslog_hostname` and `syslog_port`.

treydock:
@kibahop Thanks for the pull request, this will be released as v7.2.0

PR #192, opened 2021-04-22
Domain mode sockets

kibahop:
Be explicit about using ha sockets with domain mode


treydock:
@kibahop Thanks for the pull request. This is being released as v7.2.2

PR #197, opened 2021-07-10
Remove prefixes from socket-binding-groups

kibahop:
This fixes the problem of using wrong address for socket-binding-groups in 01-https-proxy.epp caused by misuse of prefixes. This only affected domain mode, but was also wrong, albeit working for standalone mode. 


treydock:
Ubuntu failures are long standing issue with MySQL module on Ubuntu 20.04.  Changes look good and for people using standalone (like myself), this won't actually change anything in rendered config.cli.

PR #203, opened 2021-08-12
Add extra configurations to keycloak realm

qboileau:
Add support of :
- `accessCodeLifespanLogin`
- `actionTokenGeneratedByAdminLifespan`
- `actionTokenGeneratedByUserLifespan`
- `offlineSessionIdleTimeout`
- `offlineSessionMaxLifespan`
- `offlineSessionMaxLifespanEnabled`

Tested on Keycloak 11

### DIFF
--- a/spec/unit/puppet/type/keycloak_realm_spec.rb
+++ b/spec/unit/puppet/type/keycloak_realm_spec.rb
@@ -49,7 +49,7 @@ describe Puppet::Type.type(:keycloak_realm) do
     events_listeners: ['jboss-logging'],
     admin_events_enabled: :false,
     admin_events_details_enabled: :false,
-    offline_session_max_lifespan_enabled: false,
+    offline_session_max_lifespan_enabled: :false,
   }
 
   describe 'basic properties' do
--- a/spec/unit/puppet/type/keycloak_realm_spec.rb
+++ b/spec/unit/puppet/type/keycloak_realm_spec.rb
@@ -49,6 +49,7 @@ describe Puppet::Type.type(:keycloak_realm) do
     events_listeners: ['jboss-logging'],
     admin_events_enabled: :false,
     admin_events_details_enabled: :false,
+    offline_session_max_lifespan_enabled: false,
   }
 
   describe 'basic properties' do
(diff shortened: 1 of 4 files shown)


spec/unit/puppet/type/keycloak_realm_spec.rb
```diff
@@ -49,6 +49,7 @@
     events_listeners: ['jboss-logging'],
     admin_events_enabled: :false,
     admin_events_details_enabled: :false,
+    offline_session_max_lifespan_enabled: false,
```
treydock, line 52:
```suggestion
    offline_session_max_lifespan_enabled: :false,
```

treydock: The unit tests will skip checking for default if value is `false` and technically the default is a symbol.

https://github.com/treydock/puppet-module-keycloak/blob/2e099ed88f6ec9156b192905813016d3e0890409/spec/unit/puppet/type/keycloak_realm_spec.rb#L155

treydock:
Be good to add these parameters to realm acceptance tests to validate Keycloak deployment will accept these values:

https://github.com/treydock/puppet-module-keycloak/blob/2e099ed88f6ec9156b192905813016d3e0890409/spec/acceptance/2_realm_spec.rb#L11

https://github.com/treydock/puppet-module-keycloak/blob/2e099ed88f6ec9156b192905813016d3e0890409/spec/acceptance/2_realm_spec.rb#L113

treydock:
The validation failures can be fixed by running `bundle exec rake rubocop:auto_correct` if you have Ruby environment setup. If you don't have it setup, you can just change integers in acceptance tests so this

```ruby
expect(data['ssoSessionMaxLifespan']).to eq(36000)
```

Becomes this:

```ruby
expect(data['ssoSessionMaxLifespan']).to eq(36_000)
```

Notice the underscore for integer after 3 digits.  Looks like 4 instance where large integers upset rubocop.

qboileau:
Thank you for your review. I've fixed up the commit on realm spec, this should be good now. 

Btw, I've added more token settings related parameters to realm create and update spec for completeness’s sake. 
And I've fix symbol syntax on keycloak_realm_spec defaults, like you suggested.

treydock:
This was released as v7.6.0, thanks!

PR #206, opened 2021-08-20
Fix assigning management interfaces to logical interfaces in domain mode

kibahop:
In newest Keycloaks (>10), the stock host-master.xml does not include the native interface definition in management-interfaces. Fix assign-management-interfaces-master augeas resource.


treydock:
This is tagged as v7.7.1 and will be released to Forge once the GH Actions complete.

PR #207, opened 2021-08-26
Added support for bearer-only configuration of keycloak_client

verrydtj:


treydock:
This will be released as v7.8.0 once the Github Actions complete

PR #211, opened 2021-09-03
Add feature `user_managed_access_allowed` property

rdcuzins:
keycloak_realm to accept user_managed_access_allowed

### DIFF
--- a/spec/acceptance/2_realm_spec.rb
+++ b/spec/acceptance/2_realm_spec.rb
@@ -150,6 +150,7 @@ describe 'keycloak_realm:', if: RSpec.configuration.keycloak_full do
         registration_allowed => true,
         reset_password_allowed => true,
         verify_email => true,
+        user_managed_access_allowed => true,
         access_code_lifespan => 3600,
         access_token_lifespan => 3600,
         access_code_lifespan_login => 3600,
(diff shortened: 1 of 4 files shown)


spec/acceptance/2_realm_spec.rb
```diff
@@ -37,6 +37,7 @@ class { 'keycloak':
         offline_session_idle_timeout             => 2592000,
         offline_session_max_lifespan             => 5184000,
         offline_session_max_lifespan_enabled     => true,
+        user_managed_access_allowed              => true,
```
treydock, line 40: Be good to add a validation this was set here:

https://github.com/treydock/puppet-module-keycloak/blob/e18d185b4ef4c71ba1f3d7be9eadf1aef1a75778/spec/acceptance/2_realm_spec.rb#L54

Also be useful to either move this `true` value to the update function and validate that is set there or just leave unset in update section and validate set to `false`.  Here:

https://github.com/treydock/puppet-module-keycloak/blob/e18d185b4ef4c71ba1f3d7be9eadf1aef1a75778/spec/acceptance/2_realm_spec.rb#L147

https://github.com/treydock/puppet-module-keycloak/blob/e18d185b4ef4c71ba1f3d7be9eadf1aef1a75778/spec/acceptance/2_realm_spec.rb#L191

treydock:
This will be released with v7.10.0 once Github Actions complete

PR #212, opened 2021-09-07
add saml-user-attribute-mapper support

aba-rechsteiner:


aba-rechsteiner:
@treydock please help

treydock:
The current failures are actually allowed to fail, it's just an annoying issue with Github Actions not having real way to ignore certain failures.  https://github.com/treydock/puppet-module-keycloak/actions/runs/1209212021 That shows the action as green despite those 2 failures. Those 2 failures appear to be issues with systemd module that's unrelated to your changes.  #213 should fix those failing tests.

PR #216, opened 2021-09-15
set keycloak_server in keycloak_conn_validator from 'localhost' to $service_bind_address

### LINKED ISSUE #215: Keycloak connection validator server address
Hey folks,

I want to run a standalone keycloak cluster with jdbc_ping. For this, the listen address has to be set explicitly, which causes the connection validation to fail.

A merge request for this is coming soon.

Best regards
Harm

hugendudel:
Fixes `keycloak_conn_validator` when `service_bind_address` is set to different value then `0.0.0.0` or `127.0.0.1`.  [fixes #215]

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -483,8 +483,14 @@ class keycloak (
     Class['keycloak::sssd'] ~> Class['keycloak::service']
   }
 
+  if $service_bind_address == '0.0.0.0' {
+    $validator_keycloak_server = '127.0.0.1'
+  } else {
+    $validator_keycloak_server = $service_bind_address
+  }
+
   keycloak_conn_validator { 'keycloak':
-    keycloak_server => $service_bind_address,
+    keycloak_server => $validator_keycloak_server,
     keycloak_port   => $http_port,
     use_ssl         => false,
     timeout         => 60,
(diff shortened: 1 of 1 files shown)


manifests/init.pp
```diff
@@ -484,7 +484,7 @@
   }
 
   keycloak_conn_validator { 'keycloak':
-    keycloak_server => 'localhost',
+    keycloak_server => $service_bind_address,
```
treydock, line 487: This won't work when the default is used which is `0.0.0.0`.  I think this needs to be a condition like this:

```puppet
if $service_bind_address in ['localhost','127.0.0.1','0.0.0.0'] {
  $validator_keycloak_server = 'localhost'
} else {
  $validator_keycloak_server = $service_bind_address
}
```

Then change `keycloak_server` to pull value from `$validator_keycloak_server`

hugendudel: Hello @treydock,

to be honest I had exactly this discussion with a work colleague. The address `0.0.0.0` is resolved to the loopback address (`127.0.0.1`), I couldn't find anything on this behaviour in a hurry.
To rule out possible edge cases, I would provide condition handling for `0.0.0.0` as you suggested.

Greetings
Harm

treydock:
This is set to be released with v7.9.1 once the Github Actions complete.

PR #217, opened 2021-09-21
Fix and tune mangement interface definitions for both master and slave

kibahop:


treydock:
This will be released with v7.10.0 once Github Actions complete

PR #232, opened 2022-01-11
Fix issues with install_base /opt/keycloak

dmaes:
When using my own packages, IMVHO it makes more sense to use `/opt/keycloak` instead of `/opt/keycloak-$version`. So I need to set `$keycloak::install_dir` to `/opt/keycloak`, but this conflicts with the symlink in `manifests/config.pp`, so this PR fixes that.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -2,9 +2,11 @@
 class keycloak::config {
   assert_private()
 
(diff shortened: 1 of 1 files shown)


treydock:
This will be released as v7.12.1 once Github Actions complete.

PR #236, opened 2022-02-03
Fix authorization services data corruption on unrelated client changes

mattock:
Disabling authorization services for a client results, by design, removal of
all the authorization data for that client. Before this commit any property
changes (e.g. changing root_url) to a keycloak_client would generate a payload
that did not have authorizationServicesEnabled included. Keycloak interpreted
such a payload as "disable authorization services for the client", which in
turn resulted in unintentional destruction of the client's authorization data.
This commit fixes that behavior and adds tests to prevent future regressions.


mattock:
@treydock any comments or good to merge?

treydock:
This will be released as v7.12.2 once the github actions complete

PR #241, opened 2022-04-25
Support Debian 11

### LINKED ISSUE #240: Support for Debian 11
I've tested this module on Debian 11 with Keycloak 16.1.0 and it works with no issues. It would be great if we can list Debian 11 as supported by the module.

vilhelmprytz:
fixes #240


treydock:
Turns out adding Debian 11 was rather involved since a lot of the testing in this module comes from a central repo and is synced with PDK.

treydock:
This will be released as v7.17.0 once Github Actions complete

PR #244, opened 2022-05-03
fix profile.properties file path in domain mode

surcouf:
According to official documentation, profile.properties must be placed under:

- standalone/configuration/profile.properties for standalone mode
- or domain/servers/server-one/configuration/profile.properties for server-one in domain mode.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -49,12 +49,13 @@ class keycloak::config {
       $_server_conf_dir
     ]
 
-    file { $_dirs:
-      ensure => 'directory',
-      owner  => $keycloak::user,
-      group  => $keycloak::group,
-      mode   => '0755',
-    }
+  }
+
+  file { $_dirs:
+    ensure => 'directory',
+    owner  => $keycloak::user,
+    group  => $keycloak::group,
+    mode   => '0755',
   }
 
   exec { 'create-keycloak-admin':
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -56,12 +56,18 @@ class keycloak::config {
     }
   }
 
+  file { $_server_conf_dir:
+    ensure => 'directory',
+    owner  => $keycloak::user,
+    group  => $keycloak::group,
+    mode   => '0750',
+  }
+
   exec { 'create-keycloak-admin':
     command => "${_add_user_keycloak_cmd} ${_add_user_keycloak_args} && touch ${_add_user_keycloak_state}",
     creates => $_add_user_keycloak_state,
     notify  => Class['keycloak::service'],
     user    => $keycloak::user,
-    require => File[$_server_conf_dir],
   }
 
   if $keycloak::operating_mode == 'domain' {
@@ -243,13 +249,6 @@ class keycloak::config {
     line   => "JAVA_OPTS=\"${_java_opts}\"",
     match  => '^JAVA_OPTS=',
     notify => Class['keycloak::service'],
-  }
-
-  file { $_server_conf_dir:
-    ensure => 'directory',
-    owner  => $keycloak::user,
-    group  => $keycloak::group,
-    mode   => '0750',
   }
 
   file { "${_server_conf_dir}/profile.properties":
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -38,9 +38,6 @@ class keycloak::config {
     $_add_user_keycloak_args = "--user ${keycloak::admin_user} --password ${keycloak::admin_user_password} --realm master"
     $_java_opts_path = "${keycloak::install_base}/bin/standalone.conf"
 
-    $_dirs = [
-      $_server_conf_dir
-    ]
   } else {
     $_server_conf_dir = "${keycloak::install_base}/domain/servers/${keycloak::server_name}/configuration"
     $_add_user_keycloak_args = "--user ${keycloak::admin_user} --password ${keycloak::admin_user_password} --realm master --sc ${_server_conf_dir}/" # lint:ignore:140chars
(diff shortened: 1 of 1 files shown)


manifests/config.pp
```diff
@@ -245,14 +244,14 @@
     notify => Class['keycloak::service'],
   }
 
-  file { "${keycloak::install_base}/${_subdir}/configuration":
+  file { ${_server_conf_dir}:
```
treydock, line 247: You would need to define `$_server_conf_dir` for when `$::keycloak::operating_mode != 'domain'`.

manifests/config.pp
```diff
@@ -40,7 +40,6 @@
   } else {
     $_server_conf_dir = "${keycloak::install_base}/domain/servers/${keycloak::server_name}/configuration"
     $_add_user_keycloak_args = "--user ${keycloak::admin_user} --password ${keycloak::admin_user_password} --realm master --sc ${_server_conf_dir}/" # lint:ignore:140chars
-    $_subdir = 'domain'
```
treydock, line 43: If `$_subdir` is no longer needed, would need to also remove that variable from the first part of this condition.

surcouf:
Thanks for your review: I've replace $_subdir variable by $_server_conf_dir for standalone mode.

treydock:
Unit tests are failing because now duplicate resources for the directory:

https://github.com/treydock/puppet-module-keycloak/blob/335ef6d809e3bf64550f3d2cc03797f7d0e0cb6f/manifests/config.pp#L52

https://github.com/treydock/puppet-module-keycloak/blob/335ef6d809e3bf64550f3d2cc03797f7d0e0cb6f/manifests/config.pp#L248

I think this is what needs to change:

https://github.com/treydock/puppet-module-keycloak/blob/335ef6d809e3bf64550f3d2cc03797f7d0e0cb6f/manifests/config.pp#L46-L50

So maybe try this:

```puppet
    $_dirs = [
      dirname(dirname($_server_conf_dir)),
      dirname($_server_conf_dir),
    ]
```

manifests/config.pp
```diff
@@ -244,13 +249,6 @@
     notify => Class['keycloak::service'],
   }
 
-  file { $_server_conf_dir:
```
treydock: I'd recommend keeping this resource and only using `$_dirs` for domain since that's where the path is deeply nested. Also the permissions on this resource are more appropriate for configs.

manifests/config.pp
```diff
@@ -45,15 +49,16 @@
     $_dirs = [
       "${keycloak::install_base}/domain/servers",
       "${keycloak::install_base}/domain/servers/${keycloak::server_name}",
```
treydock, line 51:
```suggestion
      dirname(dirname($_server_conf_dir)),
      dirname($_server_conf_dir),
```

manifests/config.pp
```diff
@@ -45,15 +49,16 @@
...
 
-    file { $_dirs:
-      ensure => 'directory',
-      owner  => $keycloak::user,
-      group  => $keycloak::group,
-      mode   => '0755',
-    }
+  }
+
+  file { $_dirs:
```
treydock, line 57: Per other comment, it's better to only use this `$_dirs` resource where actually needed which is when domain mode is enabled.  The standalone it's not needed since the directory is already present during install, Puppet just manages the permissions.

manifests/config.pp
```diff
@@ -245,14 +245,7 @@
     notify => Class['keycloak::service'],
   }
 
-  file { "${keycloak::install_base}/${_subdir}/configuration":
```
treydock, line 248: You'll want to keep this resource and remove `$_server_conf_dir` from `$_dirs` because this resource has more restricted permissions since it will contain configs.

manifests/config.pp
```diff
@@ -34,19 +34,19 @@
...
-    $_subdir = 'domain'
     $_java_opts_path = "${keycloak::install_base}/bin/domain.conf"
 
     $_dirs = [
-      "${keycloak::install_base}/domain/servers",
-      "${keycloak::install_base}/domain/servers/${keycloak::server_name}",
-      "${keycloak::install_base}/domain/servers/${keycloak::server_name}/configuration",
+      dirname(dirname($_server_conf_dir)),
+      dirname($_server_conf_dir),
+      $_server_conf_dir
```
treydock, line 49:
```suggestion
```

manifests/config.pp
treydock: Also you can just call this resource `$_server_conf_dir`.

treydock:
The domain mode acceptance tests are failing. It looks like you need to have this resource require `File[$_server_conf_dir]`.

https://github.com/treydock/puppet-module-keycloak/blob/335ef6d809e3bf64550f3d2cc03797f7d0e0cb6f/manifests/config.pp#L60-L65

manifests/config.pp
```diff
@@ -57,6 +56,13 @@
     }
   }
 
+  file { $_server_conf_dir:
+    ensure => 'directory',
+    owner  => $keycloak::user,
+    group  => $keycloak::group,
+    mode   => '0750',
+  }
```
treydock, line 64: I'd put this file back where it was, the order in manifest doesn't matter, though might by chance make things work, what is really needed is `require` line like one of your commits had.  Puppet does not guarantee that order in manifests is order of execution so need to set hard dependencies.

treydock: I think if you revert `dc63f7a08fa4b32750c13704455baaf83ddb715f` should be good.

treydock:
This will be released with v7.19.0 once Github Actions complete.

PR #252, opened 2022-07-13
Add support for syncResgistrations

NITEMAN:


treydock:
Be good to include this in unit tests.

Default here: https://github.com/treydock/puppet-module-keycloak/blob/b2063dcf30f1473fe36642234ed4c726924fd9f4/spec/unit/puppet/type/keycloak_ldap_user_provider_spec.rb#L171

Test value here: https://github.com/treydock/puppet-module-keycloak/blob/b2063dcf30f1473fe36642234ed4c726924fd9f4/spec/unit/puppet/type/keycloak_ldap_user_provider_spec.rb#L224

treydock:
This will be released with v8.1.0 once Github Actions complete

PR #257, opened 2022-10-17
Add support for centralized logging via Gelf

nblock:
Centralized logging via Gelf is supported since Keycloak 19+.


treydock:
This will be released with v8.3.0 once Github Actions complete

PR #263, opened 2022-10-31
Add db_collate param

NITEMAN:
Rel https://github.com/puppetlabs/puppetlabs-mysql/issues/1507

Provide a `db_collate` param mimicking current's `db_charset`


treydock:
This will get released as v8.5.0 once Github Actions complete.

Issue #274, opened 2023-02-07
remove default value for hostname / make hostname optional


usermms23:
A default value is used for the hostname parameter. That means the hostname is always set in keycloak.conf. If the parameter hostname-strict is set to false, it is necessary to remove the hostname in keycloak.conf so that keycloak responds to multiple names behind reverse proxy.

Here you can find more information: https://keycloak.discourse.group/t/keycloak-not-responding-to-multiple-names-behind-reverse-proxy/15959

treydock:
Added ability to set `hostname => 'unset'` to remove from configs to bypass default.  This will be released with v9.3.2 once Github Actions complete.

PR #288, opened 2023-05-16
Do not reassign '$hostname' variable

danifr:
Addresses issue reported in https://github.com/treydock/puppet-module-keycloak/issues/274#issuecomment-1506531875

```
Info: Retrieving plugin
Info: Loading facts
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Cannot reassign variable '$hostname' (file: /etc/puppetlabs/code/environments/keycloak_21/modules/keycloak/manifests/init.pp, line: 360, column: 17) on node XXXX.bbp.epfl.ch
Warning: Not using cache on failed catalog
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -354,16 +354,16 @@ class keycloak (
       $validator_server = $config['http-host']
     }
   } else {
-    if $config['hostname'] in ['unset', 'UNSET'] {
-      $hostname = $facts['networking']['fqdn']
+    if ! $config['hostname'] {
+      $effective_hostname = $facts['networking']['fqdn']
     } else {
-      $hostname = $config['hostname']
+      $effective_hostname = $config['hostname']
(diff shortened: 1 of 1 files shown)


treydock:
@danifr If you rebase against master branch , that should clear up the failing tests.

manifests/init.pp
```diff
@@ -354,16 +354,16 @@
       $validator_server = $config['http-host']
     }
   } else {
-    if $config['hostname'] in ['unset', 'UNSET'] {
```
treydock, line 357: This unset hack needs to stay.  I'd recommend just renaming the variable.

treydock:
Fixed in #299

danifr:
Thanks @treydock. I went on vacation and I forgot about this.

PR #304, opened 2023-10-13
supress warnings

trefzer:
Add parameters to supress warnings, since my monitoring is not happy with warnings !

### DIFF
--- a/spec/unit/puppet/type/keycloak_identity_provider_spec.rb
+++ b/spec/unit/puppet/type/keycloak_identity_provider_spec.rb
@@ -156,6 +156,7 @@ describe Puppet::Type.type(:keycloak_identity_provider) do
   describe 'boolean properties' do
     # Test boolean properties
     [
+      :no_client_secret_warning,
       :enabled,
       :trust_email,
       :store_token,
--- a/spec/unit/puppet/type/keycloak_identity_provider_spec.rb
+++ b/spec/unit/puppet/type/keycloak_identity_provider_spec.rb
@@ -156,7 +156,6 @@ describe Puppet::Type.type(:keycloak_identity_provider) do
   describe 'boolean properties' do
     # Test boolean properties
     [
-      :no_client_secret_warning,
       :enabled,
       :trust_email,
       :store_token,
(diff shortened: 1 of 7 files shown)


treydock:
What about the ability to suppress all warnings with a single property?  

https://github.com/treydock/puppet-module-keycloak/blob/master/lib/puppet/type/keycloak_api.rb

That allows setting things that can be accessed by all types/providers.

They are set here: https://github.com/treydock/puppet-module-keycloak/blob/master/lib/puppet/provider/keycloak_api.rb#L13-L22

You can then do like `if self.class.suppress_warnings` in instance methods.

treydock:
Please rebase against master branch as I've recently fixed tests.

spec/unit/puppet/type/keycloak_identity_provider_spec.rb
```diff
@@ -156,13 +156,14 @@
   describe 'boolean properties' do
     # Test boolean properties
     [
+      :no_client_secret_warning,
```
treydock: The current boolean tests don't work with `parent: Puppet::Parameter::Boolean` if I had to guess since the old way kept booleans as symbols and new way uses proper booleans.

trefzer: I remove it, and try again !

spec/unit/puppet/type/keycloak_realm_spec.rb
```diff
@@ -285,6 +285,7 @@
   describe 'boolean properties' do
     # Test boolean properties
     [
+      :no_password_warning,
```
treydock: This would also need to be removed or remove `parent: Puppet::Parameter::Boolean` from the property.

spec/unit/puppet/type/keycloak_ldap_user_provider_spec.rb
```diff
@@ -257,6 +257,7 @@
   describe 'boolean properties' do
     # Test boolean properties
     [
+      :no_bind_credential_warning,
```
treydock: This would also need to be removed or remove `parent: Puppet::Parameter::Boolean` from the property.

treydock:
If you are able to update so I can modify your fork's branch via this PR, I can make the updates.

spec/unit/puppet/type/keycloak_ldap_user_provider_spec.rb
trefzer: removed

spec/unit/puppet/type/keycloak_realm_spec.rb
trefzer: removed

trefzer:
@treydock I think at least one statement of your last comment should be negated ;)
feel free to take over but hopefully it will pass now with all the removes !

PR #307, opened 2024-04-03
Fix #306 - Retrieve parentId by realm name

### LINKED ISSUE #306: Difficulties to create an user provider for master realm
Dear maintainer,

I would like to create an ldap user provider to the master realm (to allow connecting using an LDAP account instead of local accounts).

The issue is, when I'm trying to do so, the module use the `realm` name as the `parentId` name. In this specific case, it's false, the `master` realm is created automatically by keycloak, and keycloak create an Id with some random numbers/letters.

- I would like a simple way to create an LDAP user provider for the master realm. The simplest, as a user point of view, would be the module to automatically get the Id of the realm to complete `parentId` field. It would be transparent, no need to debug and understand a specific field isn't right and that's why my LDAP user provider does not appear.

You can get the master realm id with the following command: `/opt/keycloak/bin/kcadm-wrapper.sh get realms/master | jq -r .id`. (I don't know if there is a better way to find it, but just in case here's the command).

- The simplest solution to code would be to allow manually setting a `parentId` different of the `realm` field. For the module's user it means he must find a way to automatically find the master realm ID, store it and use it when trying to create the LDAP user provider (Nothing impossible). Maybe, in this case, add a warning inside the documentation ?

Thanks for your help.

TuningYourCode:
This change will fix #306 as it will try to retrieve the uuid of the given realm first and pass it as `parentId`.

It is a quiet similar solution like in https://github.com/treydock/puppet-module-keycloak/blob/master/lib/puppet/provider/keycloak_ldap_mapper/kcadm.rb#L127

### DIFF
--- a/lib/puppet/provider/keycloak_ldap_user_provider/kcadm.rb
+++ b/lib/puppet/provider/keycloak_ldap_user_provider/kcadm.rb
@@ -57,13 +57,29 @@ Puppet::Type.type(:keycloak_ldap_user_provider).provide(:kcadm, parent: Puppet::
     end
   end
 
+  def get_parent_id(realm)
+    parent_id = nil
+    output = kcadm('get', 'realms', realm, nil, ['id'])
+    Puppet.debug("#{realm} realms: #{output}")
+    begin
+      data = JSON.parse(output)
+    rescue JSON::ParserError
+      Puppet.debug('Unable to parse output from kcadm get realms')
+      data = []
+    end
+    data.each do |d|
+      parent_id = d['id']
+    end
+    parent_id
+  end
+
   def create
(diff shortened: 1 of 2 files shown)


treydock:
@TuningYourCode Can you please rebase this PR against this repo's master branch?  There was a recently merged fix that's needed to get tests passing

treydock:
The unit tests here need updates:

https://github.com/treydock/puppet-module-keycloak/blob/179598240124315f3c0aad029827cc06e733bab3/spec/unit/puppet/provider/keycloak_ldap_user_provider/kcadm_spec.rb#L52

Might be just 

```
allow(resource.provider).to receive(:get_parent_id).with('test').and_return('test')
```

TuningYourCode:
Thanks for the fix - didn't had time yet to look at it :)

PR #312, opened 2024-06-11
Implement realm otp, webauthn, webauthn passwordless and bruteforce properties

TuningYourCode:
This PR implements all webauthn and bruteforce properties on a realm supported by at least Keycloak 22.0.0.

WebAuthn Extra Origins and passwordless WebAuthn properties are not yet supported as these were introduced in later Keycloak versions.

### DIFF
--- a/lib/puppet/type/keycloak_realm.rb
+++ b/lib/puppet/type/keycloak_realm.rb
@@ -338,6 +338,42 @@ Manage Keycloak realms
     newvalues(:true, :false)
   end
 
+  newproperty(:permanent_lockout, boolean: true) do
+    desc 'permanentLockout'
+    newvalues(:true, :false)
+    defaultto :false
+  end
+
+  newproperty(:max_failure_wait_seconds, parent: PuppetX::Keycloak::IntegerProperty) do
+    desc 'maxFailureWaitSeconds'
+    defaultto 900
+  end
+
+  newproperty(:minimum_quick_login_wait_seconds, parent: PuppetX::Keycloak::IntegerProperty) do
+    desc 'minimumQuickLoginWaitSeconds'
+    defaultto 60
+  end
+
+  newproperty(:wait_increment_seconds, parent: PuppetX::Keycloak::IntegerProperty) do
+    desc 'waitIncrementSeconds'
+    defaultto 60
+  end
+
+  newproperty(:quick_login_check_milli_seconds, parent: PuppetX::Keycloak::IntegerProperty) do
+    desc 'quickLoginCheckMilliSeconds'
+    defaultto 1_000
+  end
+
+  newproperty(:max_delta_time_seconds, parent: PuppetX::Keycloak::IntegerProperty) do
+    desc 'maxDeltaTimeSeconds'
+    defaultto 43_200
+  end
+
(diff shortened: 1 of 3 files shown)


TuningYourCode:
@treydock do you know what i did wrong or if there is some kind of debug possibility?

Properties seem to work in my own builds so i don't get why acceptance tests fail.

treydock:
The Keycloak 24.0.3 failures are the following:

```
       	Error: Parameter web_authn_policy_authenticator_attachment failed on Keycloak_realm[test]: Invalid value "cross-platform". Valid values are plattform, cross-plattform, not specified. (file: /tmp/apply_manifest_151227864.pp.V1dPa1, line: 2)
```

```
     Failure/Error: expect(data['rememberMe']).to eq(true)
       
       expected: true
            got: false
```

```
     Failure/Error: expect(names).to eq(['profile'])
       
       expected: ["profile"]
            got: ["role_list", "profile", "acr", "web-origins", "roles", "email"]
```

```
     Failure/Error: expect(data['eventsEnabled']).to eq(true)
       
       expected: true
            got: false
```

```
     Failure/Error: expect(expected_roles - realm_roles).to eq([])
       
       expected: []
            got: ["other_new_role"]
```

treydock:
For Keycloak 22.0.0 failures, it looks like something is configured in such a way to cause Keycloak to crash.

TuningYourCode:
Thanks, despite the typo i found the issue. It's related to the old version the tests are running against. 

e.g. with KeyCloak 23.0.0 introduced the WebAuthn Extra Origins property (https://www.keycloak.org/docs/23.0.0/release_notes/index.html#webauthn-improvements) which was present in my initial PR. It was working in our setup as we already run KeyCloak version 24.0.5.

So i just dropped the properties which are not supported by the tested KeyCloak versions (i anyway don't need them for now)

treydock:
You can only add these new properties that work with 24.x when that's the version being tested:

```
let(:new_properties) do
  if RSpec.configuration.keycloak_version.split('.').first.to_i >= 24
    [
       "key => 'value'",
       "anotherkey => 'anothervalue'",
    ]
  else
    []
  end
end
```

Then in the Puppet code for acceptance tests can do like:

```
#{new_properties.join("\n")}
```

I'd be fine merging with the acceptance test lines for 24.x properties commented out and work later to integrate them or just drop 22.x tests and uncomment them later.

Also need to include new properties in unit tests: https://github.com/treydock/puppet-module-keycloak/blob/master/spec/unit/puppet/type/keycloak_realm_spec.rb

TuningYourCode:
From my point of view this PR should be ready for merging now. This PR only includes settings present in at least KeyCloak 22.0.0.

There might be "breaking changes" if somebody is using `custom_properties` for properties now implemented in the resource. Not sure if you consider them to be "breaking changes" that this PR has to wait for a major release.

I also created a seperate PR (https://github.com/treydock/puppet-module-keycloak/pull/313) for raising minimal KeyCloak to version 23.0.7 (which is out of support if not the RedHat build is used - see: https://www.keycloak.org/security.html) which might definitly be a "breaking change" for a major puppet module version.

treydock:
This will be released at 11.2.0 once Github Actions complete

PR #326, opened 2024-11-12
Remove support for logging via GELF

nblock:
Logging via GELF was deprecated in Keycloak 24 and was finally removed Keycloak 26.


treydock:
I'll hold off merging this until the next major version bump is ready which will also include the default version being at least Keycloak 26.

PR #328, opened 2024-11-16
Reuse session when running kcadm commands

### LINKED ISSUE #327: Reuse existing sessions created by kcadm-wrapper.sh?
The current implementation of [`kcadm-wrapper.sh`](https://github.com/treydock/puppet-module-keycloak/blob/8516850878b9f6f6dba5dda9c11beb4e87e38542/templates/kcadm-wrapper.sh.erb) uses `--no-config` and performs a full login for each request to the admin REST API. This has a few drawbacks:

* A full login is slow, especially since Keycloak changed the settings of PBKDF2 in Keycloak 23.
* Each login creates a new session which pollutes the session log.
* With persistent user sessions (Keycloak 25+) this also consumes more resources in the database).

With this [PoC-patch](https://github.com/nblock/puppet-module-keycloak/commit/c12a91b92f2484b6f404caf806e6e1fb75f428e0) we got some improvements in our environment:

* Before: 1011.32 seconds for a Puppet run on Keycloak 26 and 760 new "admin" sessions in the database
* After: 971.41 seconds for a Puppet run on Keycloak 26 and 1 new "admin" session in the database
* NOTE: the reduced runtime is "only" ~40 seconds because we reduced the number of rounds in the hash algorithm for the master realm. If you are using Keycloak's default values, you should see a bigger speed improvement.

There might be way better ways to implement this, for example by providing the ability to overwrite the `kcadm-wrapper.sh`.

treydock:
Fixes #327


files/kcadm-wrapper.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# shellcheck source=/dev/null
+. /opt/keycloak/conf/kcadm-wrapper.conf
+
+EXPIRES=$(/usr/bin/sed  -n -r 's|.*"refreshExpiresAt" : ([0-9]*).*|\1|p' "$CONFIG" || echo "0")
+NOW=$(/usr/bin/date +%s%3N)
+
+if [ ! -f "$CONFIG" ] || [ "$EXPIRES" -gt "$NOW" ]; then
```
nblock, line 9: This should be: `if [ ! -f "$CONFIG" ] || [ "$EXPIRES" -lt "$NOW" ]; then`

PR #330, opened 2024-11-18
Rename to hide_on_login_page to hide_on_login

nblock:
It seems that Keycloak 26 deprecated hideOnLoginPage - see https://www.keycloak.org/docs/latest/upgrading/index.html#improving-performance-for-selection-of-identity-providers

Using the deprecated hideOnLoginPage yields:

```
Keycloak_identity_provider[x on y]/hide_on_login_page: defined 'hide_on_login_page' as 'true' (corrective)
```


treydock:
Unit tests need updates:

https://github.com/treydock/puppet-module-keycloak/blob/8516850878b9f6f6dba5dda9c11beb4e87e38542/spec/unit/puppet/type/keycloak_identity_provider_spec.rb#L99

https://github.com/treydock/puppet-module-keycloak/blob/8516850878b9f6f6dba5dda9c11beb4e87e38542/spec/unit/puppet/type/keycloak_identity_provider_spec.rb#L165

treydock:
Once fixed I'll merge with next major release which will include at least Keycloak 26 as the default.

PR #343, opened 2025-01-23
Add backchannel_logout attributes

wolfaba:
We have additional updates, please, consider merging.

In this PR we add backchannel_logout_session_required and backchannel_logout_revoke_offline_tokens attributes

Thank you.

Regards,
Robert.


spec/acceptance/5_client_spec.rb
```diff
@@ -11,29 +11,33 @@ class { 'keycloak':
...
-        roles                          => ['bar_role', 'other_bar_role'],
+        realm                                    => 'test',
+        root_url                                 => 'https://test.foo.bar',
+        redirect_uris                            => ['https://test.foo.bar/test1'],
+        default_client_scopes                    => ['address'],
+        secret                                   => 'foobar',
+        login_theme                              => 'keycloak',
+        backchannel_logout_url                   => 'https://test.foo.bar/logout',
+        backchannel_logout_session_required      => 'true',
+        backchannel_logout_revoke_offline_tokens => 'true',
```
treydock, line 22: For these new parameters, can you verify they are actually getting set?  Further down is where the expected values are confirmed to be set.

PR #344, opened 2025-02-04
Protocol mapper aggregate attrs

wolfaba:
We have additional updates, please, consider merging.

In this PR we add aggregate.attrs for protocol mapper.

Thank you.

Regards,
Robert.


wolfaba:
@treydock there are some errors again, but maybe something with keycloak process?

PR #360, opened 2026-04-21
feat: adding support for attributes in keycloak_client

roock:
This allows to set the attributes inside the keycloak_client, e.g. additional authentication mechanisms like oauth2 device authorization.

```json
  "attributes": {
    "realm_client": "false",
    "oidc.ciba.grant.enabled": "false",
    "backchannel.logout.session.required": "true",
    "standard.token.exchange.enabled": "false",
    "display.on.consent.screen": "false",
    "oauth2.device.authorization.grant.enabled": "true",
    "backchannel.logout.revoke.offline.tokens": "false"
  },

```

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1569,6 +1569,10 @@ access.token.lifespan
 
 adminUrl
 
+##### `attributes`
+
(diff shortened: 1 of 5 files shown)


treydock:
Released as v14.1.0

PR #365, opened 2026-08-10
Add Ubuntu 26.04 support

mancz:
Adds Ubuntu 26.04 LTS to `operatingsystem_support`.

We run this module (v14.2.0) on Ubuntu 26.04 hosts with Keycloak 26.x in production without issues — the manifests contain no OS-release-specific branches for Ubuntu, so metadata is the only change needed. Java comes in via the `java` class (OpenJDK 21 on 26.04), which resolves fine.

Happy to add an acceptance nodeset for 26.04 in a follow-up once container images for it are common in your CI setup.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -12,6 +12,7 @@ Rakefile:
       - debian-12
       - ubuntu-2204
       - ubuntu-2404
+      - ubuntu-2604
     puppet:
       - puppet8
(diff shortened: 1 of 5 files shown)


mancz:
Closing: the module works on Ubuntu 26.04 out of the box for our use case.

PR #366, opened 2026-09-20
freeipa_user_provider: add trust_email and enabled parameters, fix ensure

drkp:
freeipa_user_provider wraps keycloak_ldap_user_provider but doesn't provide all of its options. This PR extends freeipa_user_provider to pass through the trust_email (treat emails as verified) and enabled options. Both default to the same values as the underlying type, so existing manifests are unaffected

I also noticed that freeipa_user_provider has a `ensure` parameter, but always passes `ensure => present` through to keycloak_ldap_user_provider. This fixes it to pass the ensure parameter through so `ensure => absent` works correctly.


treydock:
Debian 11 failures are likely due to EOL, will resolve that in a different PR.

